PR #64, opened 2026-08-22
(fix) use the detected context window in acp_status and compress

Jstn-1g:
## Problem

On a model with a 1,000,000-token context window, the human `/acp status` command reports the expected pressure, but the model-facing `acp_status` and `compress` tools can still calculate pressure against the 128,000-token fallback. This inflates usage by about eight times and can produce false over-limit or emergency guidance.

Thank you to @adamqeqj003-dotcom for the detailed diagnosis and proposed correction in #63.

## Cause

`handleStatus` and `handleCompress` passed the startup `ToolEnvironment` directly to `kernelConfigFor`. Automatic window detection is exposed through `windowFor(agent)` and does not mutate that startup object, so these two paths never received the detected limit.

## Fix

Both handlers now:

1. ask `windowFor(agent)` for the effective model window;
2. build the kernel configuration with that detected limit;
3. preserve the existing explicit 128,000 fallback when no resolver is available.

No report formatting or compression policy is reimplemented; the existing kernel receives the correct adapter value.

## Verification

- regression tests record the configuration passed to `processTurn` and prove that both `acp_status` and `compress` receive `1,000,000`, not `128,000`
- `npm test` — 162 passed
- `npm run typecheck` — passed
- `npm run build` — passed

## Documentation review

The README configuration table already states that the automatically detected `modelContextLimit` drives kernel pressure decisions. This patch makes the two model-tool paths match that documented behavior, so no documentation text changes are needed.

### DIFF
--- a/docs/acp-status-align-design.md
+++ b/docs/acp-status-align-design.md
@@ -192,6 +192,8 @@ buildStatusReport(state: CompressionState, messages: CoreMessage[], countTokens:
 
 **tokenCount 口径（P2-2）**：`handleStatus` 调 `processTurn` 时 `tokenCount` 用 `resolveTokenCount(agent, surfaceMessages)`（surface 口径，与 `handleCompress`/`buildNudge` 一致，遵守硬性规则 2）；**不持久化 turn.state**（`env.store.set` 不调用），避免同 turn 二次推进 nudge 基线、`Nudge:` 行只用 `turn.nudge.reason`。
 
+**窗口来源（issue #63）**：`config.modelContextLimit` 不再是 `env.modelContextLimit`（128K 初始兜底），而是经 `windowFor(agent)` 探测后的有效窗口——`handleCompress` / `handleStatus` 与人类侧 `statusText` 统一走 `resolveEffectiveWindow(env, agent)`（`src/tools.ts`），再 `kernelConfigFor({ ...env, modelContextLimit: window.limit })`。模型工具只消费探测后的窗口进 nudge 决策（规则 9：窗口行不进工具输出），窗口来源的人类可见性由 `/acp` 的 `context window:` 行承担。
+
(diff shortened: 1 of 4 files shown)


Tyan66666:
Thanks for the fix — the tool-side pressure now follows the detected window. I built on top of your commit with two additions:

**1. (refactor) share the effective-window resolution across tools and /acp**

Your fix inlined the window resolution in the two tool handlers; the same inline block already existed in the `/acp` command (`statusText`). That made three copies of the same logic. This extracts a single `resolveEffectiveWindow(env, agent)` helper in `src/tools.ts` and uses it from all three call sites — compress, acp_status, and `/acp status`. Behavior is identical (probe via `windowFor` when present, else fall back to `modelContextLimit`); it just lives in one place now. The design doc (docs/acp-status-align-design.md §4.2) now records that the tools feed the probed window into the kernel config.

**2. (test) acp_status pressure follows the probed window — false-alarm regression**

Your spy test asserts the kernel receives the probed window. This adds the user-visible regression from issue #63: with a 500-token fallback and a probed 1M window, `acp_status` must report `Nudge: idle —`, not ACTIVE. Before the fix the kernel got the 500 fallback, so a healthy session read ~25× over-limit and alarmed. The existing "small window" ACTIVE test acts as the control — without `windowFor` the fallback path still injects, so the fallback behavior is not accidentally broken.

Verified: `npm run typecheck && npm test && npm run build` — 163 tests green (162 baseline + 1 new). The false-alarm test fails on the pre-fix code (the kernel reads the fallback window → ACTIVE).

A separate follow-up is being tracked for the gateway case in #63 where auto-detection itself fails (pressure falls back to 128K until you set `modelContextLimit` explicitly); that is a probe-path concern, not a tool-path one, so I kept it out of this PR.

PR #77, opened 2026-08-25
(feat) guide the model to drill into acp_status before compressing when unsure

Tyan66666:
## Problem

The nudge already told the model to re-run acp_status when seqs go stale (the range table is a snapshot), but nothing told it to check when it is *unsure about context pressure or block structure*. In practice the model compresses straight from the nudge table without ever looking at the status report — it can miss signals only acp_status shows (block sizes, pressure breakdown, block ledger).

## Cause

Both places that describe acp_status — the system-prompt quick-reference and the tool description — framed it purely as a stale-ref refresher, never as an uncertainty check.

## Fix

Add a short push to both spots: before compressing, if you are unsure about context pressure or block structure, run acp_status to drill in first.
- src/prompts.ts: system-prompt acp_status entry + tool description gain the uncertainty trigger
- tests/prompts.test.ts: byte-identical default-description snapshot updated

## Verification

- typecheck passes
- 174/174 tests pass (full suite)
- build succeeds


Tyan66666:
已并入 #73（acp-index 分支）：三处 drill-in-first 引导（系统提示条目 + acp_status 工具描述 + nudge footer）随 #73 一起发布，避免两个 PR 各带一半引导。

PR #118, opened 2026-09-06
(fix) page large-block decompress output with offset/limit so one recovery can't flood or get silently trimmed

### LINKED ISSUE #112: [bug][recall] decompress / /acp decompress 输出无上限、无分页 —— 大 block 恢复要么灌爆上下文,要么被宿主静默截断
来源: ranxianglei/billion-context#582 (仿照 billion-context-pi#305 的 bug 历史对照扫描) 分析 decompress 召回路径时发现。

## 现象
`decompress({ blockId })` 与 `/acp decompress <blockId>` 把目标 block 展开后的**全部** shadowed 原文一次性拼进单个 tool result / command 返回文本,没有任何 `limit` / offset / 大小上限:

- `src/tools.ts:549-572` (`handleDecompress`):`expandShadowedSeqs(session, block.blockId)` 展开全部 seq(含 tier-2/3 递归到 parent 原文),逐条 `extractEventText` 后 `[...].join('\n\n')` 整体返回。参数 schema 只有 `blockId`(`src/tools.ts:528-530`)。
- `src/commands.ts:115-131` (`decompressText`):同一模式,同样无界。

## 影响
1. **大 block 恢复直接灌爆上下文**:一个覆盖数百 KB 工具输出的 T1 block(或 distill 了多个 T1 的 T2/T3 block)decompress 一次就把几十万 token 打回模型上下文——恰好抵消 ACP 压缩刚省下的空间,可能立刻把 usage 推过 emergency 线,触发 #108 描述的 emergency nudge 重注入路径,形成"解压→爆→再压"震荡。
2. **被宿主静默截断时返回无用残片**:宿主侧 dsh-compaction-tool-result-pruner 对 tool result 有 8192/4096/1024 档裁剪(见 billion-context-pi#305 floor 4 的核对记录);插件侧没有任何自己的 gate 或截断标记。大 block 恢复结果很可能只剩前几 KB,而模型不知道后面还有内容、也不知道如何继续取(没有 offset 语义)。
3. `search_context` 只给 160 字符 preview(`src/tools.ts:647 previewLength: 160`),模型判断"值得完整恢复"之后唯一的通道就是这个无界一次性恢复。

对照:billion-context 的 decompress 有 `toFile` 逃生通道(恢复内容落盘、上下文里只留摘要行,配合 #112 临时文件 reaper),从机制上避免恢复即灌爆;DSH 没有文件系统通道,所以更需要分页。

## 建议修复
- 给 `decompress` 增加可选 `offset` / `limit`(按 shadowed seq 顺序分页),`acp_status` drilldown 或 search_context 命中信息里带上 block 原文总 token 数,让模型先估量再决定恢复范围;
- 或至少加硬上限 + 显式截断 footer(如 `… truncated at seq N of M — call decompress again with offset=N`),保证模型拿到的是"可继续的截断"而不是"静默残片";
- `tests/tools.test.ts` 现有 decompress 用例只验证恢复正确性(如 151/176/945 行附近),补一个超大 block 的分页/截断行为用例。

ranxianglei:
Fixes #112.

**Problem**
`decompress` (and `/acp decompress`) expands ALL shadowed seqs of a block into one uncapped tool result. For a large block (say 200K tokens) the recovered text either floods the model's context or — more usually — gets silently truncated by the host tool-result pruner (`dsh-compaction-tool-result-pruner`, tiers 8192/4096/1024): the tail of the recovered content is gone while the model believes it holds the whole block. That is precisely the failure mode decompress exists to undo.

**Cause**
`handleDecompress` (src/tools.ts) joins `expandShadowedSeqs(session, block.blockId)` into a single string; the tool schema only exposes `blockId`, so there is no way to page through.

**Fix**
- `sliceDecompressPage(expanded, offset, limit)` + `DEFAULT_DECOMPRESS_PAGE = 100` (src/region.ts): pure paging helper shared by the tool and the command. Clamps `offset >= 0`, `limit >= 1`; empty-text seqs still occupy slots so paging stays stable across calls.
- `decompress` tool: optional integer `offset` / `limit` params. Results end with `[messages X..Y of Z]` plus an explicit continue hint (`decompress({ blockId: "<id>", offset: N })`) when more remains; an offset past the end returns guidance instead of an error.
- `/acp decompress <blockId> [offset] [limit]`: same paging and footer (8-char prefix form); non-numeric args are rejected.
- Default prompt copy (src/prompts.ts) + tool descriptions updated to mention paging; README.md / README.en.md tool table synced.

Small blocks behave as before (a `<= 100` message block returns whole, plus the new `[messages X..Y of Z]` footer).

**Verification**
- New regression tests:
  - tests/tools.test.ts "M3: decompress pages a large block by default and walks it with offset" — compresses 120 messages, asserts page 1 = `[messages 1..100 of 120]` + continue hint, walks to the final page via offset, and checks past-end guidance.
  - tests/tools.test.ts "M3: decompress honors an explicit limit and clamps a negative offset".
  - tests/commands.test.ts "M4: /acp decompress pages with [offset] [limit] args and rejects bad numbers".
- `npm run typecheck` clean; `npm test` 211 pass / 0 fail (208 baseline + 3 new); `npm run build` clean (dist committed per AGENTS.md §4).

Built against main @ 6348af8 (v0.2.20).


ranxianglei:
我来协助看一下这个问题,请分析并回复处理结果。

Tyan66666:
## Merge fix pushed

I merged current `main` into this branch (`cee6ca1`) and fixed the two things the merge broke:

- **`tests/tools.test.ts`** — both new paging tests called `rebuildBlockLedger(session.events)`. `main` moved the seam to `session.snapshotEvents()` (dsh-session 0.1.5), so after the merge they failed with `TypeError: 'events is not iterable` (at `src/region.ts:486`). Changed to `session.snapshotEvents()`.
- **Dropped the branch's `dist/` changes.** `dist/` is now refreshed by `.github/workflows/dist-bot.yml` after every merge, and PR branches are expected to carry whatever `dist/` main last had (`AGENTS.md:153`; the CI drift check no longer gates PRs — `ci.yml:66` is schedule/bot-only). Your PR body's "dist committed per AGENTS.md §4" was correct at base `6348af8`; the rule changed under you, nothing to fix on your side.

Gates on the merged tree: `npm run typecheck` clean, `npm test` 245/245, working tree clean.

## Review of the paging design

The direction is right — one uncapped `decompress` result was the bug, and paging is the shape of the fix. But as written the branch does not yet close the second failure mode in #112, and one line of the copy can be lost by the very mechanism the PR is about. Line numbers below are from the branch **after** merging `main`.

### 1. The page is budgeted in messages; the host trims in characters — so the default page is still silently trimmed

`DEFAULT_DECOMPRESS_PAGE = 100` (`src/region.ts:1127`) and `safeLimit = Math.max(1, Math.floor(limit))` (`src/region.ts:1149-1151`) bound *message count*. The host pruner is a **character** budget, and this repo already documents its numbers: `docs/dsh-porting-analysis.md:55` — `dsh-compaction-tool-result-pruner` | `thresholdChars 8192 / head 4096 / tail 1024`.

Measured against the PR's own fixture (`tests/helpers.ts`, `LONG` = 206 chars; `tests/tools.test.ts:195` uses `buildTextSession(121)`): one message renders to ≈224 chars on the page, so a 100-message page is ≈22,400 chars — **2.7× the 8192 threshold**. A page only stays under it below ≈36 such messages. So the default page still gets silently trimmed, which is exactly #112(b).

Two follow-ups worth deciding together:

- Size the page by accumulated characters (stop at a documented budget below 8192, leaving room for the header/hint and the `[seq N]` prefixes), keep `offset` as the message index for continuation; or derive the budget from the kernel's token counter. Either way the constant should be a *size* budget.
- Which part of an oversized result does the host actually drop? This PR's description says the tail is gone; `docs/dsh-porting-analysis.md:55` says head 4096 + tail 1024 are **kept** and the middle is dropped. I could not settle it from the repo: `dsh-compaction-tool-result-pruner` is not in `node_modules` (a grep finds only `@deepseek-ai/dsh-compaction/README*.md`), so the implementation isn't reachable locally. Worth confirming on a live host — the fix is the same, but the wording in the PR body is not.

### 2. `limit` has no ceiling

`Math.max(1, Math.floor(limit))` clamps the floor only, so `decompress({ blockId, limit: 100000 })` returns the whole block and re-opens the flooding half. Suggest a maximum, plus `minimum`/`maximum` in the schema at `src/tools.ts:536-537` so a bad value fails loudly instead of silently clamping.

### 3. A non-numeric `offset` leaks `NaN` into the user-facing message

`Math.max(0, Math.floor("abc"))` is `NaN` → `expanded.slice(NaN, NaN)` is `[]` → the model sees `offset NaN is past the end`. Reachable through the wrapped `{ arguments: { … } }` envelope, which bypasses the top-level schema (that envelope is live-verified; `AGENTS.md` rule 9). A `Number.isFinite` guard fixes it.

### 4. The "more available" hint sits at the tail of the payload

`src/tools.ts:588-590` and `src/commands.ts:155-160` append `[messages X..Y of Z]` and the continue hint **after** the content. If the page is oversized, the host drops the middle of the result — so the one line that tells the model the page is partial can be exactly what disappears, and a partial page reads as complete. Put the marker and the hint before the content (a trailing copy too is fine).

### 5. Sizing should follow a kernel-owned budget (rule 7)

The kernel owns truncation sizing and its marker: `config.truncate.threshold` (`acp-kernel/dist/index.js:352`), `TRUNCATION_MARKER` (`index.js:538`), `truncateLargeToolOutputs` (`index.js:545`). There is no kernel pager — I checked, nothing paging-shaped in `acp-kernel/dist/*.d.ts` — so hand-rolling the pager is fine and is not a rule-11 workaround. But a private message-count constant plus private footer copy sits on the wrong side of that line; the *budget* should come from a size budget (chars) or the kernel's counters.

### 6. Docs not synced (§4 requires it in the same PR)

- `docs/configurable-prompts-design.md:342` quotes the old default `decompress` description verbatim while `src/prompts.ts:210` now appends the paging sentence — and that doc promises its examples track the source character-for-character (`:347`).
- `docs/INSTALL.md:164` — the manual verification step (`decompress({ blockId })`) should mention paging.
- `AGENTS.md` §2 module map, the `region.ts` row — the new exports (`sliceDecompressPage`, `DEFAULT_DECOMPRESS_PAGE`) belong there.

### 7. Missing regression test

Nothing in the suite turns red if a page exceeds the host's 8192-char budget, so #112(b) is currently unguarded. A test with a multi-KB message — ideally a real nested `tool-result` fixture, which is also the tool-heavy case the paging tests don't cover — asserting the emitted page stays under that budget would close it. The existing `M3: decompress pages a large block by default and walks it with offset` (`tests/tools.test.ts:195`) already guards the off-by-one arithmetic, which is good.

### 8. Wording

`README.md:191`, `README.en.md:198`, `src/prompts.ts:210,239` and the PR title promise a result that "can never … get silently trimmed by the host". That isn't true for a 100-message page today — either size the page (item 1) or soften the claim.

## Smaller things, no urgency

- The two entry points disagree on bad input: the tool silently clamps, `/acp` rejects (`src/tools.ts:574-575` vs `src/commands.ts:132-134`). Also the tool's hint prints the full UUID while `/acp` prints the 8-char prefix.
- `DecompressPage.limit` (`src/region.ts:1120-1121`) is returned but never read.
- The doc comment claims page boundaries are "stable across calls" (`src/region.ts:1143-1147`), but `expandShadowedSeqs` is recomputed per call — a `compress` between two pages shifts them. True only while the log is frozen.
- A direct unit test for `sliceDecompressPage` (floor clamp, no ceiling, `NaN`, `offset > total`, `exhausted`) would pin the helper itself rather than only its callers.

Happy to implement any of this on the branch if you'd rather not — the merge fix is already pushed either way.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework-aio)

Implemented all of the design feedback on the branch — pushed as `45a9e51` to the PR head (fast-forward from your merge `cee6ca1`). Item by item:

**1. Page budgeted in chars, not just messages.** Added `DEFAULT_DECOMPRESS_PAGE_CHARS = 7000` (`src/region.ts`) alongside the message count. `sliceDecompressPage` now bounds each page by **both** ≤100 messages **and** ≤7000 rendered characters, pricing each message exactly as it appears on that page (the tool path includes the `[seq N] ` prefix). 7000 sits below the host pruner's 8192 threshold with headroom for the header, prefixes, and continue hint, so an ordinary page comes back intact — this closes #112(b). The message count doubles as the hard ceiling (see #2).
- On *which* part the host drops (your PR body said tail; `docs/dsh-porting-analysis.md:55` says head+tail kept, middle dropped): I couldn't settle it from the repo either — `dsh-compaction-tool-result-pruner` isn't in `node_modules`, only its READMEs. The fix is identical either way (keep the page under the threshold), and the leading-marker change (#4) means the "this page is partial" line survives whichever part gets trimmed. I kept the copy honest ("a normal page comes back intact") rather than claiming it can never be trimmed.

**2. `limit` ceiling.** Capped at `DEFAULT_DECOMPRESS_PAGE` (100) in the slicer, so `limit: 100000` returns at most 100 messages instead of the whole block. One constraint worth flagging: the `dsh-tools` schema DSL has **no** `minimum`/`maximum` keys — `IntegerValueSchemaSpec` only supports `type`/`enum`/`const` — so the cap can't be expressed in the schema itself. It's enforced in code and documented in the param description ("values above 100 are capped to 100"). `/acp` rejects out-of-range loudly (human-facing).

**3. Non-numeric `offset`.** Guarded with `Number.isFinite`; non-finite or non-numeric `offset`/`limit` fall back to defaults instead of leaking `NaN` into the slice. This absorbs the wrapped `{ arguments }` envelope form too.

**4. Hint at the tail.** The `[messages X..Y of Z]` marker and continue hint now **lead** the payload in both `handleDecompress` (`src/tools.ts`) and `/acp decompress` (`src/commands.ts`), so the partial-page line isn't the exact line a middle-trim would drop.

**5. Kernel-owned budget (rule 7).** Deliberately did **not** tie this to acp-kernel's `config.truncate.threshold`: that knob truncates a single oversized tool output during compression, whereas the host pruner trims our whole decompress result — different mechanisms, different thresholds. Hand-rolling the pager stays fine (no kernel pager exists), and the constant is now a size budget (chars), not a bare message count.

**6. Docs synced.** `docs/configurable-prompts-design.md` decompress quote re-synced verbatim to the new `src/prompts.ts` copy; `docs/INSTALL.md` step 4 mentions paging; the `AGENTS.md` module-map `region.ts` row notes the new exports; both READMEs updated.

**7. Regression tests.** Added:
- A nested `tool-result` multi-KB fixture (real `helpers.appendToolResult` shape) that walks a large block and asserts **every emitted page is < 8192 chars** — the #112(b) guard that was missing.
- Direct `sliceDecompressPage` unit tests pinning the helper itself: 100-message ceiling binding, oversized-limit clamp, char-budget stop, `NaN`/bad-input fallback, `offset > total`, and the single-huge-message "always include the first" forward-progress guarantee.
- Updated the existing walk/limit tests for the new sizing.

**8. Wording.** Softened "can never … get silently trimmed" to "a normal page comes back intact" across `README.md`, `README.en.md`, and `src/prompts.ts`.

Smaller things: kept the intentional tool-clamps-vs-`/acp`-rejects split (added a comment explaining why); corrected the "stable across calls" doc comment to "while the log is frozen"; left `DecompressPage.limit` in the return shape for API stability.

Gates on the pushed commit: `npm run typecheck` clean, `npm test` 247/247, `npm run build` ok. `dist/` intentionally not committed (dist-bot refreshes it after the merge). Note the PR currently reads "behind" main because main advanced past your `cee6ca1` merge — worth a fresh rebase/re-merge before landing so CI runs green.

修了什么问题：decompress 分页原来只按消息条数限页，而宿主 pruner 按字符（8192）截断，长消息宽页仍会被从中部裁掉、且续页提示恰在被裁的尾部——现改为「条数 + 字符预算」双限（每页 ≤100 条且 ≤7000 字符）、给 limit 加上限、加 NaN 防护、把标记/续页提示移到结果开头，并补齐了 <8192 的回归测试与文档同步；改动已推到 PR 分支，测试全绿，可合并。

ranxianglei:

Follow-up to my last note — I went ahead and re-merged main myself rather than leaving the branch behind.

While pushing I found main had advanced past the branch base (`215a54c`) with two commits I didn't have: **`d63317c`** (fix #93 — stop compressing injected AGENTS.md rows, break the compress→re-inject loop, protect the real last user turn) and its dist refresh. That's what had flipped the PR to a conflicted state.

Re-merged them in (`dc5f43f`, pushed to the PR head):
- Only **`AGENTS.md`** conflicted — both sides had extended the same `region.ts` module-map line. Resolved by keeping **both** notes (decompress paging #112 + instruction-row barriers rule 16). Every code file auto-merged cleanly.
- Verified the combined tree: `npm run typecheck` clean, `npm test` **261/261** (the extra ~14 are main's new `instruction-barrier.test.ts` coverage running alongside my paging tests).

The PR now reports `mergeable: true` — no conflict. It shows `blocked` only because the required CI checks (`ci` + `pr-title`) are still running against the new head `dc5f43f`; that clears on its own once CI goes green. I haven't merged it (human-only per repo rules).

把分支重新合并了最新 main（#93 注入指令行修复 + dist 刷新），唯一冲突的 AGENTS.md 按「两边都保留」解决，合并后树 typecheck 干净、261/261 测试全绿，PR 已可合并（当前 blocked 只是等 CI 检查跑完，会自动转绿）。

PR #119, opened 2026-09-06
(fix) cap emergency nudge injections per user turn (closes #108)

### LINKED ISSUE #108: [compression] emergency nudge 无上限逐 step 重注入，且缺少 compress attempt 断路器
来源: ranxianglei/billion-context-pi#305 分析 "对照本项目分析billion-context-dsh中存在的bug" 时发现

## 现象
当上下文压力持续处于 emergency 阈值（默认 `0.85`）以上时，每个 LLM step 都会重新注入一条完整的 emergency nudge 消息，且该消息是持久追加到会话日志的（会被 append 成 `user/message` 事件）。一旦压力降不下来（模型反复 no-op / 选错范围 / 内容大部分被保护），nudge 本身会进一步抬高压力，形成正反馈环。

## 证据
- `src/nudge.ts:144` `const emergency = nudge.breakdown?.emergencyOverride === 1`
- `src/nudge.ts:147` `const alreadyShown = !emergency && lastNudgeTurn.get(session.id) === turnNumber` —— emergency 直接 bypass per-turn dedup
- `src/nudge.ts:148` `if (alreadyShown) return null`（emergency 恒不命中）
- `src/index.ts:308` `return { kind:'enter', messages:[...decision.messages, outcome.message] }` —— 每 step 追加一条持久消息
- `src/index.ts:168` 默认 `nudgeEmergencyThresholdPct: 0.85`
- 全仓 grep 无任何 attempt 计数 / 断路器 / cooldown，只有 per-turn dedup（`lastNudgeTurn`）
- `tests/nudge.test.ts:83-93` 明确断言 "emergency nudges bypass the per-turn dedup"（有意设计但缺上限）
- no-op 结果 `Compressed 0 block(s)`（`src/tools.ts:402-409`）不 terminal，不会停掉后续 nudge

## 影响
token 泄漏 + 面板刷屏。对照 billion-context-pi 的同类失败模式：#223（transient retry prompt 每次调用重注入，实测 ~400 次/小时、紧急 pct 95→127%，最终整体移除）、#250（dead-range compress loop breaker）、commit `414acd1`（cap emergency nudge injections per user turn）。dsh 三者皆无。

## 建议修复
1. emergency nudge 也加 per-user-turn 注入上限（对齐 pi commit `414acd1`）
2. 引入 compress attempt 断路器：同一 user turn 内连续 N 次（建议 3）failed/no-op 后停止重注入，仅保留 kernel 机械兜底
3. no-op 结果计入失败预算

ranxianglei:
Closes #108.

## Problem

When context pressure stays above the emergency threshold (engine default `0.85`), **every** agent step re-injects a full emergency nudge, and each nudge is appended as a durable `user/message` event. If the pressure does not come down (the model no-ops, picks the wrong range, or the content is protected), the nudge's own tokens push usage even higher — a positive feedback loop that never stops. billion-context-pi hit the same failure mode (#223: retry prompt re-injected ~400×/hour, eventually removed) and fixed it with a per-user-turn cap (commit `414acd1`).

## Cause

`src/nudge.ts` dedups only **normal** nudges per user turn (`lastNudgeTurn`); the emergency path bypassed the dedup entirely (`!emergency && …`). No attempt counter, cap, or cooldown existed anywhere in the repo.

## Fix

- `EMERGENCY_NUDGE_MAX_PER_TURN = 3` (hardcoded — a safety mechanism, not a tuning knob): emergency nudges are capped at 3 injections per user turn, tracked in a per-session `emergencyNudges` map (`{ turn, count }`). The budget resets when the user turn advances (same semantics as the normal dedup).
- Normal nudges are unchanged: still at most one per user turn, and a normal nudge does NOT consume the emergency budget.
- When the kernel still wants an emergency nudge but the budget is spent, the engine logs a `warn` (`emergency nudge suppressed — per-turn budget of 3 spent …`) so operators can see WHY the model stops receiving nudges (issue #108 review).
- Docs updated in the same PR: README zh/en config table, config comments, AGENTS.md module map.

**Scope note:** this PR implements proposal #1 from the issue (the per-turn cap), which bounds the unbounded re-injection loop — the token-leak part. The compress-attempt circuit breaker (proposals #2/#3) is a follow-up; the kernel-side `minPressureBenefit` gate lands with the acp-kernel ≥0.0.54 bump planned in #116.

## Review follow-up (2026-09-08)

Applied from the review on #108:

- Cap regression test rewritten with explicit per-call assertions: calls 1–3 inject, **call 4 returns null**, later calls stay null; the cap-hit hook fires for every suppressed call.
- New test: the cap resets on a new user turn.
- New test: same-turn normal + emergency mix — a normal nudge does not consume the emergency budget.
- `logger.warn` on cap hit (in the `agent/pre-step` handler, `src/index.ts`).
- Cap stays hardcoded at 3 (not exposed as a config knob).
- Branch rebased onto current `main` (v0.2.21 + e2e harness); no `dist/` changes carried — per the current policy the dist-bot refreshes `main` after the merge.
- Title renamed to the `(fix) …` form required by the pr-title CI check (the old `fix(nudge): …` form fails `scripts/check-pr-title.mjs`).

## Verification

- `npm run typecheck` — clean
- `npm test` — **216/216 pass** (including the three #108 tests above)
- `npm run build` — clean (local check only; `dist/` not committed)


ranxianglei:
我来协助看一下这个问题,请分析并回复处理结果。

PR #121, opened 2026-09-06
(fix) bound per-session kernel-state and nudge-dedup caches with LRU eviction

ranxianglei:
Fixes #113.

**Problem**
Two per-session caches grow without bound for the lifetime of the engine process:
- `AcpStateStore.states` (src/state.ts) — one full acp-kernel `CompressionState` per session id, including every block summary; `delete()` had zero callers.
- `lastNudgeTurn` (src/index.ts) — one entry per session id that ever received a nudge.

A long-lived server handling many sessions leaks memory proportional to total historical session count × per-session state size. (`compressCallIdsToHide` is intentionally left alone: it is bounded by in-flight compress calls and self-cleans on tool/result arrival.)

**Cause**
Plain `Map`s with only insert paths; nothing evicts idle sessions.

**Fix**
- New `src/lru.ts`: `LruMap<K,V>` (extends `Map`, API-compatible drop-in) + `DEFAULT_SESSION_CACHE_LIMIT = 512`. Recency refreshes on both get and set; over-cap inserts evict the coldest entry.
- `AcpStateStore.states` is now an LRU map capped at 512 sessions (constructor takes an optional limit for tests).
- `lastNudgeTurn` is now an LRU map with the same cap. Worst case after eviction of a cold session: one extra non-emergency nudge when it returns (per-turn dedup restarts) — bounded and self-healing.
- **Why eviction is lossless for block identity:** `stateFor` already rehydrates from the durable log (the same path a restart uses), and rehydration is deterministic — bN ids are recorded in the durable event or synthesised in ledger order, and message refs are re-derived from the append-only full-log array (`allLogMessages` includes shadowed originals), so numbering reproduces exactly.
- **Secondary fix found while making eviction routine:** rehydration set `blocks`/`nextBlockId` but not `nextRunId`, so the first compression after a restart (or eviction) reused run id `r1` colliding with a rehydrated block. Rehydration now continues run ids after the rehydrated max (`nextRunIdAfter`).
- Kernel fields that do reset on eviction — `tokenSnapshot`, nudge cadence state, stats counters — all self-heal on the session's next turn; documented on the field.

**Verification**
- New tests:
  - tests/lru.test.ts (4): cap enforcement, recency refresh on get AND set, delete semantics, cap clamping. The re-set test caught a real bug in the first implementation (`Map.set` on an existing key does not move it to MRU position — fixed by delete+re-insert).
[...]

PR #122, opened 2026-09-06
(chore) bump acp-kernel to 0.0.63 — range table from the kernel (drift workaround removed)

### LINKED ISSUE #116: [kernel] acp-kernel 钉在 0.0.29,落后主仓 25 个版本(0.0.54):应急 nudge 缺 minPressureBenefit 闸门等行为差异,#38 workaround 需在 0.0.54 上复查
来源: ranxianglei/billion-context#582 第二轮扫描(用户指定"内核他也没更新,结合内核继续扫描")。对 acp-kernel 0.0.29 vs 0.0.54 的 dist 做了逐模块 diff(npm pack 双版本对照)。

## 事实
- 本仓 devDependencies 钉 `acp-kernel@0.0.29`(精确 pin,符合纪律);billion-context 主仓当前钉 **0.0.54**(npm latest stable)。落后 25 个版本。
- **对 dsh 现有用法无 breaking change**(已验证:pipeline/state/sync/refs/report/tokenize/compression-rules/config 的 .d.ts 逐字节相同;boundaries/types 纯增量;`renderNudgeText(decision, prompts?)` 签名不变)。**默认 prompts 与 nudge 配置两版逐字节一致** —— 所以 nudge.ts 里对内核渲染文本的正则手术(replaceRangesStr/replaceTierTrigger/replaceEmergencyExample)今天没有被打穿,但这是巧合性安全:内核一旦改 prompt 文本,dsh 的正则静默失效(无测试护栏,见 #111 同族的格式耦合风险)。

## 0.0.54 的行为差异(dsh 升级即生效,需评估)
1. **`decideNudge` 新增 `minPressureBenefit` 闸门**(v29 无此函数):pressure/emergency 注入条件从 `bestPending > 0` 变为 `bestPending >= minPressureBenefit`,默认 `max(5000, 1%×contextWindow)`。这直接缓解 #108 的"emergency 无上限重注入"死循环中"无可压内容仍反复注入"分支(剩余 pending < 5K 时不再注入);但模型持续压缩失败且 pending ≥ 5K 时循环仍在,#108 的断路器诉求不变。参考实现见 bc#182/#370(同参数失败短路)。
2. **新增 `firstSightMassReady` 触发器**:会话从未 nudge 过(lastNudgeShownTokens===0 且 baseline===0)时,usage ≥ minContextLimitPct(默认 0.45)且 max pending ≥ nudgeGrowthTokens 立即就绪——大初始上下文的会话不再等增长才触发。升级后行为变化,需回归。
3. **T2 触发条件加入计数维度**:`t2Count >= tiers.tier2Trigger`(默认 5)或原 token 阈值,二者满足其一。
4. 新能力模块(absorb 即时工具结果吸收、viable ranges、handoff、parse-compress-input、panel、persist)均为 additive;其中 `parseCompressInput`/`parseCompressArgs` 与 dsh 自研的 `unwrapCompressArgs`+`validateContentItems`(src/tools.ts:222-290)功能重叠,升级后可考虑收敛到内核单一实现。

## 配套事项(按 docs/upstream-tracker.md 维护规则)
- **#38 ref-map 漂移 workaround**(`buildCompressibleSeqRanges` 自算 range 表):tracker 标注每次 kernel bump 必须复查,上游修了就删 workaround。本次 diff 未在 dist 层面看到明确修复证据,**需要在 0.0.54 上用 #38 的复现场景实测**;另 #38 E 节自认"尚未提交上游 issue",应在 ranxianglei/acp-kernel 补提。
- **#46** 依赖的 acp-kernel PR #123(reverse+offset)基于 v0.0.38,距今已隔 16 个版本,存在 stale/conflict 风险,bump 前需确认其状态。

## 建议
按 AGENTS.md §4b SOP 规划一次 0.0.29→0.0.54 的受控 bump:先跑全量测试 + e2e,重点回归 nudge 行为(minPressureBenefit/firstSightMassReady/T2 计数)、#38 workaround 是否可撤、nudge 正则手术对新内核输出仍命中的断言(建议把内核渲染文本的关键段落固化进 tests/kernel-upstream/,消除巧合性安全)。

ranxianglei:
**Update (2026-09-12): the pin goes to 0.0.63, and the engine's own range scan is gone**

I pushed the bump further, because the reason this PR existed — the ref-map drift that forced the engine to compute its own range table — is fixed upstream now.

- **Pin: `0.0.54` → `0.0.63`.** 0.0.63 contains that fix: acp-kernel#207 (`Range segmentation assumes refNum order == message array order — surface-replace hosts get end<start ranges + spurious fragmentation`) → #208/#209 (`segment compressible ranges by array adjacency, not ref arithmetic`), released in **v0.0.59**. Everything the 0.0.54 write-up below lists (first-sight gate, benefit floor, duplicate-range guard) is still in this line.
- **`src/region.ts`:** `buildCompressibleSeqRanges(session, kernelView, opts)` now translates the kernel's `compressibleRanges` into live surface seqs and applies the host-side guards on top (instruction-row barrier, checkpoint/system nodes, tail + real-user protection). The self-computed scan and its `UPSTREAM:` workaround note are deleted — geometry from the kernel, guards from the host.
- **`src/nudge.ts`:** the range table is fed `nudge.compressibleRanges` through a small `kernelRangeViewOf(nudge, state)` helper instead of re-deriving geometry.
- **e2e fixtures:** warm filler enlarged (`count: 110` → `250`). The benefit floor described in item 2 below, plus `minCompressRange` (5000 chars), silently suppressed the nudge in the harness window of 2000 — the old fixture's best pending range was 2554 tokens against a 5016-token floor — so four nudge assertions failed with no engine error to look at. That is fixture sizing, not engine behavior; both floors are ~4% of the default 128K window.
- **New test:** `tests/kernel-range-source.test.ts` (real kernel: refs → live seqs, shadowed spans never offered, a mid-span compression leaves no hole, the instruction-row barrier still bites), mutation-checked — deleting the barrier, or reverting the translation to ref arithmetic, makes it fail.
- **Kept from this PR:** the `tests/config.test.ts` / `tests/helpers.ts` / `tests/state.test.ts` updates below (adapted to main's newer `buildNudge` signature and `snapshotEvents()`), so nothing of the original work is lost.
- **Verification now:** `npm run typecheck` clean, `npm test` **299/299**, `npm run test:e2e` 4 scenarios green. `dist/` is not part of this PR — the dist bot rebuilds it after the merge.
- Current `main` is merged in (the branch had drifted): the conflicts were `AGENTS.md` and the two test files whose call sites main renamed when it moved nudge guidance out of the frame — resolved by keeping main's copy and re-applying the new signature. The push is a fast-forward, no force.

*The sections below are the original analysis, written for the 0.0.54 bump — the numbers in them (208/208 tests, dist size) describe that state, not this one.*

---

Fixes #116.

**Problem**
`acp-kernel` was pinned at 0.0.29 while the shared kernel line has moved to 0.0.54 — 25 releases of fixes and anti-thrashing gates left on the table (the pin is exact, per AGENTS.md §4b).

**Compatibility analysis (verified against both versions before bumping)**
- `.d.ts` diff: pipeline / state / sync / refs / report / tokenize / config modules byte-identical; boundaries and types additive only. No breaking changes for dsh's usage surface (`createCore`, `processTurn`, `decideNudge`, `CompressionState`, `CoreMessage`).
- Default prompts + nudge config strings byte-identical → dsh's regex-surgery adaptation layer (src/nudge.ts) is safe against 0.0.54 text as-is; covered by tests/nudge.test.ts.
- New additive kernel modules (absorb / viable / handoff / parse-compress-input / panel / persist) are simply not imported by dsh yet — no action needed now.

**Three behavior changes in 0.0.54 that touch dsh (all intentional upstream improvements)**
1. **firstSightMassReady** (kernel index.js `decideNudge`): a FRESH nudge state (never shown, no baseline) whose usage is already ≥ `minContextLimitPct` (kernel default 0.45) with mass-ready pending content fires an efficiency nudge immediately instead of waiting for growth. Effect in dsh: big fresh sessions get their first compression guidance sooner.
2. **minPressureBenefit floor**: nudges are suppressed when the best compressible range would reclaim < max(5000, 1% × window) tokens — anti-thrashing; partially mitigates #108 (emergency spam from tiny-benefit ranges).
3. **Duplicate-range guard**: a compress whose range is fully covered by active block(s) now fails with an explicit "already covered by bN" message instead of landing a duplicate block. dsh production flows are unaffected — `buildCompressibleSeqRanges` already excludes shadowed seqs, so the model can only propose ranges it reports as compressible; the guard is defense-in-depth (and closes a #120-style loop vector).

**Test updates (both encode pre-0.0.54 kernel behavior)**
- tests/config.test.ts "a lowered over-limit threshold triggers the nudge": the 61%-usage fresh session now fires under DEFAULTS via firstSightMassReady (previously growth-gated → null). Assertion inverted accordingly; each `buildNudge` call gets its own `lastNudgeTurn` map so the per-turn dedup can't mask the second scenario.
- tests/state.test.ts topic test: its second ("plain") session shared id `test-session` with the first; with the new duplicate-range guard the second compress (same store key, same covered seqs) correctly lands no block and the ledger assertion crashed. Fix: distinct session id via a new optional `id` param on `buildTextSession(count, id = 'test-session')` (backward-compatible; all other call sites unchanged).

**Verification (§4b gates)**
- `npm run typecheck` clean; `npm test` 208/208 pass; `npm run build` clean (dist committed; 194.87 KB → 212.87 KB, expected — new kernel modules inline).
- Hotspot re-checks: CJK estimator test green (1 char/token unchanged); CompressionState rehydration tests green; #38 ref-map-drift regression (tests/region.test.ts) green; nudge-text surgery tests green.
- Independent of #118 and #121 (no overlapping hunks except unrelated appends in tests/state.test.ts).


ranxianglei:
我来协助看一下这个问题,请分析并回复处理结果。

Tyan66666:
Thanks — I picked this up and pushed the bump further, because the reason this PR existed (the ref-map drift that forced the engine to compute its own range table) is fixed upstream now.

**What changed since your commit**

- The kernel pin goes to **0.0.63** instead of 0.0.54. 0.0.63 carries the upstream fix for the drift: ranxianglei/acp-kernel#207 (`Range segmentation assumes refNum order == message array order — surface-replace hosts get end<start ranges + spurious fragmentation`) → #208/#209 (`segment compressible ranges by array adjacency, not ref arithmetic`), released in **v0.0.59**.
- `src/region.ts`: `buildCompressibleSeqRanges(session, kernelView, opts)` now translates the kernel's `compressibleRanges` into live surface seqs and applies the host-side guards on top (instruction-row barrier, checkpoint/system nodes, tail + real-user protection). The self-computed scan and its `UPSTREAM:` note are gone — geometry from the kernel, guards from the host.
- `src/nudge.ts`: the range table is fed `nudge.compressibleRanges` through a small `kernelRangeViewOf(nudge, state)` helper instead of re-deriving geometry.
- Your `tests/config.test.ts` / `tests/helpers.ts` / `tests/state.test.ts` edits are kept (adapted to main's newer `buildNudge` signature and `snapshotEvents()`), so nothing from the original bump is lost.

**One fixture-size fix worth knowing about**

0.0.54+ (and 0.0.63) suppress a nudge when the largest pending range is below a benefit floor: `minPressureBenefitTokens = max(5000, round(window * 0.01))`, plus `pendingByTier` drops ranges under `minCompressRange = 5000` chars. On the e2e harness (window 2000, warm filler `count: 110`) the biggest pending range was 2554 tokens, so the kernel returned `shouldInject = false` silently and four nudge assertions went red with no engine error to look at. The fixtures now use `count: 250`, which clears the floor — a fixture-sizing fix, not an engine change (the floors only bite in toy windows; at the default 128K window they are ~4% of it).

**Verification**

- `npm run typecheck` clean
- `npm test` → 299/299
- `npm run test:e2e` → 4 scenarios, all checks green
- New `tests/kernel-range-source.test.ts` runs against a real kernel (refs → live seqs, shadowed spans never offered, a mid-span compression leaves no hole, the instruction-row barrier still bites) and is mutation-checked: deleting the barrier, or reverting the translation to ref arithmetic, makes it fail.
- `dist/` is untouched in this PR, so the dist bot can refresh it after the merge as usual.

I also merged current `main` into the branch (it had drifted): the conflicts were `AGENTS.md` and the two test files whose call sites main renamed when it moved nudge guidance out of the frame — resolved by keeping main's copy and re-applying the new signature. The push is a fast-forward, no force.

Issue #111, opened 2026-09-06
[test][caching] 缺跨轮出站字节稳定性 / prompt-cache 回归测试


ranxianglei:
来源: ranxianglei/billion-context-pi#305 分析 "对照本项目分析billion-context-dsh中存在的bug" 时发现

## 现象
dsh 用 seq-as-ref、不给消息打 tag，结构性规避了 pi 的 tag `tokens=` 漂移→前缀缓存失效一类 bug（pi #171/#173）。但仓库里没有任何跨轮出站字节稳定性 / prompt-cache 回归测试来锁死「两轮之间前缀不变」这一缓存友好性质——代码演进时可能被悄悄破坏而无告警。

## 证据
- grep `tests/` 与 `src/`：无 cross-turn prefix-stability 或 prompt-cache 断言
- 仅见 `tests/prompts.test.ts` 的跨*版本*字节快照（version anchor）与 `windowCache`/snapshotEvents 缓存测试，均非跨轮前缀稳定性

## 附带：两处摘要节点之外的 mid-prefix durable 改写（额外缓存失效点，建议一并记录）
- `hideCompressToolPair`（`src/region.ts:631-668`）：每次成功 compress 后用 microtask 把已发送的 tool-call+result 两节点 replace 成一个 user message，改写已发送字节 → 下一次请求从该位置起 cache miss（一次性、为绕 issue #18 的 400，属必要失效点）
- `stripOrphanedSurfaceToolMessages`（`src/region.ts:683-773`）：每个 pre-step 无条件跑（`src/index.ts:301`）；健康会话是 no-op，但一旦发生崩溃残留/broken pair 就会改写 mid-prefix 已发送字节 → 非预期缓存失效

## 建议
补一个多轮 prefix-stability 回归测试（对齐 pi #175 的 prefix-stability 思路）；并在文档中显式记录上述两个 mid-prefix 失效点，避免后续误以为「除压缩外前缀恒定」。

ranxianglei:
我来协助看一下这个问题,请分析并回复处理结果。

Tyan66666:
Closing: the byte-stability guard shipped in #126 (merged as `e149da9`).

`tests/byte-stability.test.ts` pins the three properties this issue asked for:
- re-deriving an unchanged session is byte-identical (the outbound request bytes of the same session must not drift);
- an append-only turn preserves every previously-sent byte — the prefix property, i.e. the cache-costly failure mode;
[...]

PR #126, opened 2026-09-06
(test) pin cross-turn outbound byte-stability / prompt-cache guard (#111)

ranxianglei:
## What
Adds the missing regression test for issue #111 — nothing currently guards that the **outbound prompt stays byte-stable across turns**, which is precisely what makes provider prompt-caching hit.

## Why it matters
The model-visible payload each turn is `eventsToCoreMessages(surfaceEventsOf(session))`. Providers only extend a cached prefix when the bytes of previously-sent content are unchanged. If any code path re-prices, reorders, re-references, or otherwise mutates earlier surface nodes when a new turn lands, every later request silently falls back to a full uncached re-bill — the same class of regression pi caught in #175. dsh had no automated guard for this.

## What's tested (`tests/byte-stability.test.ts`, unit level; host round-trips are #120's e2e scope)
1. **Determinism** — projecting an unchanged session twice yields identical bytes (no hidden timestamp / map-order / re-pricing dependence).
2. **Append-only prefix stability** — after appending one user+assistant exchange with no compression, every previously-sent message is byte-identical at its index and the sent view only grows — the exact invariant prompt-caching relies on.

Compression deliberately rewrites the surface, so it is intentionally *not* asserted here (documented in the file header).

## Verification
- `npm run typecheck` ✅
- `npm test` → **210/210** pass (main baseline 208 + 2 new); both byte-stability tests green
- Test-only change: no `src/` or `dist/` touched → CI dist-drift check unaffected

## Files
- `tests/byte-stability.test.ts` (+57): two focused guards


ranxianglei:
我来协助看一下这个问题,请分析并回复处理结果。

Tyan66666:
Pushed `a4068c4` — your branch merged with current `main`, plus the changes below. Thanks for taking #111 on; the goal is right and the fixture you built (a long first turn pushing past the kernel's compression threshold, then two short turns) is a good one.

## The problem with the current oracle

`tests/byte-stability.test.ts` compares `JSON.stringify(eventsToCoreMessages(surfaceEventsOf(session)))`. That is the projection this engine hands to `acp-kernel` (`src/messages.ts:178`) — it is **not** what a provider receives. A provider keys its prompt cache on the JSON body the agent loop sends, which the loop builds from `session.deriveMessages()` (`@deepseek-ai/dsh-agent-loop/lib/index.js:1204` → `buildRequest`). The two shapes are different: the kernel projection is `{ id, role, contentType, toolName?, toolCallId?, text }`, while the outbound messages are raw event data (`{ id, role, content, source, … }` — see `@deepseek-ai/dsh-session/lib/types/surface.js:75`).

So a change in outbound serialization — key order, spacing, an added field, a re-serialized content block — would leave the test green. That is exactly the failure mode #111 is about, and it is the same trap `docs/e2e-harness-design.md` describes as the "honest-usage projection-anchor trap".

## What the branch does now

**A. `tests/byte-stability.test.ts`** snapshots `session.deriveMessages()` — the exact message array the agent loop puts into the request body — and asserts the leading/append-only facts against it. Synthetic fixture, no harness needed, same intent as the original.

**B. `scripts/e2e/`** now keeps the raw request body (`fake-llm.mjs` used to discard it) and asserts byte-level stability at the wire:

- raw envelope — everything before `"messages"`, i.e. key order and spacing — identical across a scenario's LLM calls;
- the `tools` array byte-identical;
- the leading (system) message byte-identical from request 2 on — request 1 may legitimately precede the one-time ACP guidance injection;
- in scenarios without compaction, every earlier message stays byte-identical in the next request (append-only), which is what catches an in-place rewrite of the cached prefix.

## Verification

- `npm run typecheck` clean, `npm test` 264/264, `npm run test:e2e` PASS with 54 checks (was 41).
- Mutation-checked, so the new checks are not vacuous: mutating the second request body's tail plus `messages[1]` fails only the append-only check (`request #2 message 1`); mutating the raw body's head plus `messages[0]` fails the envelope and leading checks in all four scenarios. Both restored byte-identical afterwards.
- Docs synced in the same commit: `README.md`, `README.en.md`, `AGENTS.md`, and a new 2026-09-11 entry in `docs/e2e-harness-design.md`.

## One open point for you

A and B overlap in intent. If you would rather keep only one of them: B is the stricter guard (it sees the bytes a provider sees), while A still earns its place because it runs without the harness and pins the boundary on a synthetic fixture. Happy to drop either half if you prefer — just say which.

Tyan66666:
Two notes on the required checks, both unrelated to the code:

**1. `pr-title` is failing on this PR.** That check reads the title only, so it fails regardless of the diff — and a squash merge turns the title into the commit subject on `main`, which is why the repo constrains it (`CONTRIBUTING.md`, `AGENTS.md` §4). `test(caching): …` is not one of the accepted forms. `(test) pin cross-turn outbound byte-stability / prompt-cache guard (#111)` would pass. Accepted forms: `(feat|fix|refactor|test|chore) <summary>`, `docs: <summary>`, `release vX.Y.Z` (regex: `scripts/check-pr-title.mjs`).

**2. `ci` and `e2e` are green on `a4068c4`** — the code side is fine, so the title is the only thing standing between this PR and a merge.

Rename it yourself or say the word and I'll do it — either works.

Tyan66666:
Done — renamed to `(test) pin cross-turn outbound byte-stability / prompt-cache guard (#111)`. Revert it if you'd rather phrase it differently; the only hard requirement is the bracketed kind prefix.

ranxianglei:

Renamed title verified: I ran the repo's own linter on `(test) pin cross-turn outbound byte-stability / prompt-cache guard (#111)` → `[pr-lint] OK`, and the PR is `mergeable_state: clean`. It reads fine as-is (kind prefix + what + why + issue ref), so no need to revert.

**On the A-vs-B open point: keep both.** I pulled `a4068c4` and read the two halves rather than reason from the description, and they guard genuinely different seams — neither is vacuous given the other:

- **A** (`tests/byte-stability.test.ts`) pins the `session.deriveMessages()` seam in-process: re-deriving an unchanged session is byte-identical, and an append-only turn leaves every earlier message index byte-identical. It runs in plain `npm test` (no harness), fails fast, and its failure message points at the exact index (`outbound message N changed`).
- **B** (`scripts/e2e/run-e2e.mjs` → `cachePrefixChecks`) pins the *serialized wire body*: the raw envelope (key order + spacing before `"messages"`), the `tools` array, the leading system message from request 2 on, and the append-only prefix. Those four live outside `deriveMessages()` — they're assembled by the agent loop's `buildRequest` — so A structurally cannot see them. That is precisely the blind spot the original oracle had.

The overlap is only in *intent* (both want "earlier bytes don't change"); the artifacts differ (derived message objects vs. the HTTP bytes a provider actually keys its cache on). Dropping B reintroduces the envelope/tools/framing gap; dropping A loses the fast, harness-free unit signal, the precise per-index diagnostics, the seam-level contract doc, and any coverage when e2e isn't run.

One thing I specifically liked: A's third test still pins the old `eventsToCoreMessages(surfaceEventsOf(...))` projection, but under an explicit "not the cache boundary" name so the two can't blur together again — that directly inoculates against the exact mistake the first version made.

Nothing blocking. Required checks (ci + e2e) are green on `a4068c4`, and your mutation check confirms the new assertions aren't vacuous. Docs are synced in the same commit. Ready for a human merge — I won't merge it myself (merges are human-only).

中文摘要：把「出站字节稳定性 / prompt-cache 守卫」的断言从内核投影挪到了真正的出站层——A 在单测里钉住 `deriveMessages()`（无 harness、失败能定位到具体下标）、B 在 e2e 里钉住线上请求体（envelope 键序/空格、tools、首条系统消息、前缀），两者断言对象不同、互不冗余，建议都保留；标题已改并通过 lint、required checks 全绿，可以合并。

PR #129, opened 2026-09-06
(feat) one-word nudge presets via config.preset (five tiers)

### LINKED ISSUE #105: Feature Request: Presets + Custom UI for billion-context-dsh
Problem:
Currently, tuning the plugin requires editing YAML/prompts manually, which is intimidating for new users.

Proposal:
Add 5 presets (preserve / relaxed / balanced / efficient / aggressive) that control nudge frequency/thresholds, plus a 6th "custom" mode exposing key knobs in the UI:

- nudgeMinContextLimitPct
- nudgeMaxContextLimitPct
- nudgeEmergencyThresholdPct
- protectedLastMessages
- growthRatio
- editable system/context/turn prompts (with reset button)

Why:
Lowers the barrier, makes behavior predictable, yet keeps full flexibility for power users. All parameters already exist in the config – just need UI exposure.

Thanks!

---

**Maintainer note (2026-09-06, triage):** tracked as a sub-issue of #75 (runtime settings integration). Phase mapping and corrections:

- `nudgeMinContextLimitPct` / `nudgeMaxContextLimitPct` / `nudgeEmergencyThresholdPct` already exist in the engine config; #75 Phase 1 makes them hot-reloadable (host `settings.yaml` + `/acp config` list/set/reset — no restart, no manual YAML editing).
- `protectedLastMessages` and `growthRatio` do **not** exist in the current engine config or in `acp-kernel` — adopting them means adding new knobs first, and any kernel-level concept needs an upstream `acp-kernel` change first (this repo ports the kernel, it does not fork it).
- Editable system/context/turn prompts = #75 Phase 2 (to be designed separately).
- The graphical UI half = #75 Phase 3. Update (2026-09-06): on the DSH 0.1.2 line the `WEB_SETTINGS_NAMESPACES` gate is gone — `SettingsController.describe()` now lists every registered namespace dynamically, and the client plugin system (`dsh.client` manifest + `./client` export) lets a plugin contribute its own settings card via the `settings.plugin.item` slot. No upstream permission is needed; the remaining work is authoring our own card UI (feature-gated to the 0.1.2 line — on older hosts the custom mode stays fully served by `/acp config` + `settings.yaml`).
- The 5 presets (preserve / relaxed / balanced / efficient / aggressive) are the new scope this issue tracks: preset tiers map onto the settings keys introduced by #75.

Full roadmap lives in #75: Phase 1 (settings plumbing) → Phase 2 (prompts) → presets + custom UI (this issue) → Phase 3 (web settings card — gate lifted on the DSH 0.1.2 line, remaining work is our own client card).

ranxianglei:
## Problem

Changing how eagerly ACP nudges compression meant hand-tuning three percentages (`nudgeMinContextLimitPct` / `nudgeMaxContextLimitPct` / `nudgeEmergencyThresholdPct`) by hand. Issue #105 asks for a one-word choice instead.

## What this does

Adds an optional `config.preset` key that accepts five tiers:

| preset | min | max | emergency | trade-off |
|---|---|---|---|---|
| `preserve` | 0.55 | 0.78 | 0.93 | keep context as long as possible |
| `relaxed` | 0.50 | 0.75 | 0.90 | light-touch compression |
| `balanced` | 0.45 | 0.70 | 0.85 | **== current out-of-box defaults** (choosing it changes nothing) |
| `efficient` | 0.40 | 0.60 | 0.78 | trim more often |
| `aggressive` | 0.30 | 0.50 | 0.70 | lean context |

Design points:
- **Precedence: explicit value > preset > engine default.** The preset fills ONLY the three nudge thresholds you did not set explicitly, so a partial override on top of a preset is honored.
- **No other knob is touched** — `modelContextLimit`, `autoNudge`, `prompts`, `coreOverrides` are unaffected; `coreOverrides.nudge` still lands last.
- **Fail-fast:** an unknown name throws at engine construction and lists the valid tiers; it never silently falls back to defaults.
- **Display-only in `/acp status`:** prints the active preset and its resolved thresholds (including any explicit override you made on top).

## Scope notes

- Landed at the **composition-config layer** today (install / `cordis.patch.yml` `config:`). Exposing `preset` through settings.yaml hot-swap rides on #75 Phase 1 (`src/settings.ts` is not merged to main yet) — a small follow-up once it lands.
- `growthRatio` (kernel `nudge.growthRatio`, already reachable via `coreOverrides`) and `protectedLastMessages` (≈ kernel `preserveRecentMessages`) are **deliberately not baked into the presets** — neither is a first-class engine knob here; adopting them as named keys / UI items is an owner decision. Flagged in `docs/presets-design.md`.

## Verification

- `npm run typecheck` clean
- `npm test` → **227 pass / 0 fail** (13 new tests in `tests/presets.test.ts`; includes an end-to-end proving `efficient` fires the over-limit nudge while `balanced` stays quiet at the same ~61% usage)
- `npm run build` OK (tsup bundles `src/presets.ts`)
- Docs synced in the same change: README.md / README.en.md (config-table row + Presets section), docs/INSTALL.md, AGENTS.md module map, new docs/presets-design.md, docs/README.md index

Fixes #105.


Tyan66666:
Merged `origin/main` into the branch so the PR is mergeable again, then pushed the follow-ups the review asked for as `931b5ee`.

**Merge (conflict resolution)**
- `src/commands.ts` — kept both sides' imports (main added preset-aware lines, this branch added the settings/probe imports).
- `README.md` / `README.en.md` — kept main's emergency-threshold row and added this PR's preset row.
- `AGENTS.md`, `docs/INSTALL.md`, `docs/README.md`, `src/index.ts`, `src/tools.ts` — auto-merged.
- `dist/` is left at main's state (per `AGENTS.md`: PR branches do not carry it; dist-bot rebuilds after merge).
- Net diff vs main: 11 files, +634/−1.

**Fixes**

1. **An inverted window is now rejected at construction.** The kernel tolerates it — `validateConfig` only `console.warn`s ("Thresholds may not fire correctly"), it never rejects the config — so `preset` plus a single explicit override could silently ship an inverted window, e.g. `preset: 'preserve'` (min 0.55) + `nudgeMaxContextLimitPct: 0.5`: the over-limit line sits *below* the emergency line, the emergency path fires first, and the tier means the opposite of its name. `resolveAcpConfig` now validates the merged triple (`assertNudgeThresholdOrder`) and throws with the three resolved values. Only values that are actually set are compared, so an omitted `min` keeps falling back to the kernel default (0.45) rather than being mirrored in this repo.
2. **Two claims were factually wrong.** `src/presets.ts` and `docs/presets-design.md` both said the kernel's `validateConfig` *rejects* a reversed window — it warns. Both now say so and point at the engine's guard instead.
3. **`/acp status`'s comment overstated what the panel can see.** That line prints the composition-time `env` values, so a same-name key in `coreOverrides.nudge` — which lands inside `kernelConfigFor` — is not reflected; the printed thresholds can be *lower* than the ones actually in force. Comment, README (both languages) and the design doc now state the limitation. Closing the gap properly (reading `coreOverrides.nudge` in the panel) is a display change, so I left it to you.
4. **The mid-session test's name and comment were wrong.** It claims "an aggressive-tier preset makes the nudge fire where the default tier stays quiet" while comparing `balanced` (max 0.70) against `efficient` (max 0.60). Renamed and re-commented; the assertions are unchanged.
5. **One existing test asserted a window the new guard rejects** — `resolveAcpConfig({ preset: 'aggressive', nudgeMaxContextLimitPct: 0.9 })`, where 0.9 is above `aggressive`'s emergency 0.7. It now uses 0.6, which still demonstrates that an explicit value outranks the preset.

**New coverage (5 tests, all in `tests/presets.test.ts`)**
- Literal 15-value snapshot of the table — the README documents those numbers as the contract, and the existing invariant/monotonic tests would all survive a silent tweak.
- `coreOverrides.nudge` outranks the preset in `kernelConfigFor` (the precedence the README documents).
- An unknown preset fails at **engine construction**, not only in `resolveAcpConfig` — that is the boundary a composition row hits.
- The inverted-window rejection, plus a legal override on the same preset still resolving.
- The `/acp status` `preset:` line verbatim, including the three resolved percentages.

**Verification** — `npm run typecheck` clean; `npm test` **279/279**. Two mutations, each turning exactly one new test red: disabling the guard (`assertNudgeThresholdOrder` → no-op) fails only the inverted-window test; changing `aggressive.min` from `0.30` to `0.35` fails only the snapshot test.

Docs were updated in the same batch: both READMEs gained the inverted-window note, a note that **`max` is the value racing the host compaction-basic 80% line** (the `preserve`/`relaxed` emergency values sit above it and only act as a label upgrade), a note that `min` currently has no runtime effect, and a note that a mistyped `preset` takes the whole profile down (the bundle row deliberately carries no `config`, so the key comes from the user's own row).

ranxianglei:

Re-reviewed after your follow-up push (`931b5ee`). All claims check out, and I landed one fix directly on the branch for the gap you handed off.

**Verified independently** (checked out `931b5ee`, ran the full suite):
- `npm run typecheck` clean; `npm test` **279/279 pass** — matches your report exactly.
- Diff vs main (`baa8834`) is clean: 11 files, +634/−1, no `dist/`, every line on-topic. Confirmed `baa8834` is the true main tip and an ancestor of the branch, as you described.
- Code read-through: precedence (explicit > preset > default) correctly reads explicit values from the caller's config rather than the merged base (which would mask the preset); construction-time fail-fast verified at both boundaries (`resolveAcpConfig` and `new AcpCompactionEngine`); the third pairwise check in `assertNudgeThresholdOrder` is genuinely needed for the no-max case (e.g. min 0.9 + emergency 0.5); `kernelConfigFor` merge order confirms `coreOverrides.nudge` lands last; both READMEs, INSTALL, and the design doc match the code, including the corrected "kernel warns, doesn't reject" claims.

**Fixed directly on the branch** (the display gap you left to me), commit `3972f5f`:
- The `/acp status` preset line now mirrors `kernelConfigFor`'s merge order: a same-name key in `coreOverrides.nudge` shows through instead of being masked by the lower composition-time value. It reads `env.coreOverrides?.nudge` with a per-key fallback to the env value, so keys absent from the override still print the preset-resolved numbers.
- Added a regression test pinning the single-key override shape real composition rows use. (My first fixture spread the full kernel defaults into the override, and the panel correctly printed the overridden min/emergency too — implementation was right, fixture was unrealistic.)
- Synced the three doc spots that stated the old limitation (README.md / README.en.md bullets, presets-design.md §5 + revision record R1).
- Re-verified after the change: typecheck clean, **280/280** tests, build OK. PR is mergeable.

**Minor, non-blocking** (pre-existing, not introduced by this PR): a manual config *without* a preset that sets `nudgeMaxContextLimitPct < 0.45` with no min still slips past the guard — min stays undefined, the kernel's 0.45 default applies at runtime, so the growth trigger dies while the over-limit path still fires (graceful degradation). Left as-is per your documented choice; worth a separate issue if you ever want it covered.

Ready for merge whenever you are (merge stays human-only per repo rules).

---
中文摘要:复核确认了合并后分支的全部修复(typecheck 干净、279/279 测试通过、diff 干净无 dist);我直接在 PR 分支上补了 `/acp status` 预设行不反映 `coreOverrides.nudge` 同名键覆盖的显示缺口(commit `3972f5f`,含回归测试和三处文档同步,改后 280/280 全绿),另有一个此前就存在、本 PR 未引入的边角情形(max<0.45 且未设 min 时守卫跳过)仅作记录不阻塞——可以合并。

Tyan66666:
Re-merged `main` (it now carries the acp-kernel bump) and pushed the follow-ups that bump forced: `e211518`.

**Merge — four conflicts**

- `src/index.ts`, `src/tools.ts` — import blocks on both sides; kept the union.
- `README.md`, `README.en.md` — this branch added the `preset` row, while main had rewritten the same table with the hot-reload markers. Kept main's rows and re-inserted the `preset` row, now marked composition-only: `src/settings.ts` has no `preset` key (an unknown key is rejected), so only a composition row can set it.
- `AGENTS.md`, `docs/INSTALL.md`, `docs/README.md`, `src/presets.ts`, `src/commands.ts`, `tests/presets.test.ts` — auto-merged.
- `dist/` left at main's state (PR branches do not carry it; the dist-bot rebuilds after merge).

**The kernel bump invalidated one test premise (fixed here)**

`main` now carries #122, which moved the pinned `acp-kernel` from 0.0.29 to **0.0.63**. On that kernel a *first-sight* nudge state fires as soon as usage reaches `minContextLimitPct` (`acp-kernel/dist/index.js:2746`, `firstSightMassReady`; it also requires pending content at or above the growth floor, and the reason string is tagged `[first-sight mass]`).

That broke the end-to-end case in `tests/presets.test.ts:241`: it asserted `balanced` stays quiet at ~61% usage on a **fresh** state, but 61% clears `balanced`'s min (0.45), so the first-sight branch fires and the `max` line this test exists to observe is never reached. Measured with a throwaway probe: `balanced` fires on call 1 and is quiet on calls 2–3; `efficient` fires on all three.

Fix: each env is warmed with one call whose result is discarded, and the assertion runs on the following call — so the over-limit line (`max`) is what decides. I also passed the required fourth argument (`emergencyNudges`); the previous three-argument call never failed typecheck because `tsconfig.json` includes only `src/**`, so `tests/` is not typechecked.

**Two README claims the new kernel contradicts**

- "`min` currently does not affect runtime decisions — the kernel only reads it in `validateConfig`" is no longer true. `minContextLimitPct` is read at runtime in two places: the first-sight nudge above, and `tierCountUsageFloor` (`acp-kernel/dist/index.js:2750-2751`), which gates the T2/T3 block-count triggers. The bullet now states what `min` actually does — which is also the mechanism this test has to warm past.
- "each tier's trigger point is decided by `max`" → true for the recurring over-limit nudge, not for the one-time first-sight nudge.

**Verification** (branch `e211518`): `npm run typecheck` clean; `npm test` **318/318** (299 on main + 19 here); diff vs main is 11 files, +669/−1, every line on-topic; `dist/` byte-identical to main. `docs/presets-design.md` §7 carried stale numbers (279 total / 18 new) and is refreshed, with the conflict resolution and the warm-up rationale recorded.

PR #132, opened 2026-09-08
feat: injection governance — summary source framing, unchanged-injection elision, verified readings, slim nudge

TheTouYu:
Six additive changes on the injection / context-cost side, from an over-engineering review of our own harness runs (a 10–15 tool-call task that burned 238 calls / 296 K output tokens, largely through injected frames the model could neither date nor trust).

## What changes

- **B1 — summary source framing.** Compaction summaries re-injected into the model are prefixed with `[模型自写摘要 · 非用户原文 · 其中义务句需复核]` (model-written summary · not user text · obligation sentences need re-checking). `withSummaryFramePrefix()` is idempotent and gated by `isCompactionCheckpoint()` (`source.plugin === 'compact'`); `prefixSummaryBlocks()` applies it at creation time in `region.ts`.
- **B2 — unchanged-injection elision.** `src/injection-dedupe.ts` collapses a re-injected `agent-instructions` / `skill-catalog` / `runtime` / `runtime-context` frame to a 38 B `[unchanged: sha1:<12hex>]` stub **only when another visible copy of the same kind + hash still exists on the session surface** (`visibleInjectionKeys()` walks `session.surface.nodes` × events). When the previous copy has just been shadowed by a compaction the frame is re-injected in full — the only copy must never be deleted. No cross-request state, no reset hook.
- **B3 — verified readings.** `compress` accepts `verifiedReadings: string[]` per range; they are persisted on the `compaction/summary` event and recovered with `verifiedReadingsOf(event)`. Absent → no key written.
- **B4/B5 — state face.** Pending items carry `revokeIf`; the injected state face only carries group progress, the current step, `revokeIf`, and the next step, hard-truncated at 300 UTF-8 bytes (`FACE_BUDGET`).
- **B6 — nudge slimmed.** `stripNudgeGuidance()` removes the philosophy / how-to-compress / tier-2 / tier-3 sections from the nudge body on both the default and the template path (measured 6.1 KB → 224 B); the same guidance stays in the system prompt where it belongs.

## Tests

- `tests/injection-governance.test.ts` (8 cases: B1 idempotence + non-checkpoint pass-through, B2 visible-key semantics, B2-5 shadowing regression lock, B6 body budget).
- `tests/g5-governance.test.ts` (7 cases: B3 round-trip, missing key, non-summary events, B3-4 declaration-face lock, B6-1..B6-4).
- `tests/prompts.test.ts` updated for the new `compress` description.

## Verification (rebased on `b3a6918`)

- `npm test` → **231 pass / 0 fail**
- `npm run build` → OK (`dist/index.js` 200.42 KB)
- `npm run test:e2e` → **PASS**, all four scenarios, including `nudge injected at least once — nudges on requests 4,5,6,7`, `summary shadows first user turn`, `durable replace node landed`, and `strict tool pairing in final request` — i.e. B1/B2/B6 do not disturb projection, shadowing or message pairing.

No change to the compaction algorithm itself. B2 only elides frames whose identical content is provably still visible on the surface.

/cc thanks for `dsh-web-tools` — it replaced a broken playwright-core path in our environment with a zero-dependency CDP one.


Tyan66666:
Thanks for this — B1 and B6 are genuinely useful and I'd like to land them. B2 and B3 need work before merge, and there are three blockers that would break things on the current `main`. Review is against `origin/main` @ `baa8834` and your head `a1af8ae`; every claim below has a file:line.

## Blockers

**P0-1 — B2 is a silent no-op on the current seam.** `src/injection-dedupe.ts:81` reads `const events = session?.events ?? []`. `@deepseek-ai/dsh-session` 0.1.5-rc.2 has no `events` getter — the public surface is `eventAt(seq)` / `snapshotEvents()` (`node_modules/@deepseek-ai/dsh-session/lib/types/index.d.ts:180-196`; the class keeps only `private eventsSnapshot`). `src/session-events.ts:32-43` exists exactly so the engine never touches `.events`. So `visibleInjectionKeys()` returns the empty set forever: every injection is re-sent in full, the feature never fires, and nothing errors. Fix: `sessionEventsOf(session)` / `eventAtOf(session, seq)`.

**P0-2 — the PR's own tests crash on the current seam.** `tests/injection-governance.test.ts:56`, `:73`, `:84` use `session.events[…]`; `:143` and `:157` call `visibleInjectionKeys(session)` on a real `Session`. `tsconfig.json:19-20` excludes `tests/` from `npm run typecheck`, so this is invisible statically and fails at `npm test` with `Cannot read properties of undefined`.

**P0-3 — the top-level ledger write re-bricks pre-upgrade v0 logs (a regression of #141).** `src/region.ts:445-464` writes `tier`, `kernelBlockId`, `topic`, `parentBlockIds`, `directMessageIds`, `effectiveMessageIds`, `verifiedReadings` as top-level members of `compaction/summary`. `main` moved all of that into `rawOutput: encodeAcpBlockLedger(ledgerPayload)` (`origin/main:src/region.ts:447`) precisely because the frozen v0 reader validates an exact member allow-list and throws on the first non-admitted member. Merged as-is, every session that ran a compaction before the host upgrade becomes unopenable again — the exact bug #141 fixed. Fix: extend `AcpBlockLedgerPayload` (`src/block-ledger.ts:37-50`) and the `encodeAcpBlockLedger` / `decodeAcpBlockLedger` pair (`:64` / `:85`), keep `rawOutput`, and read `verifiedReadings` back off the decoded payload.

## B2 partly cancels #93, which is now on main

#93 landed and owns `classifySurfaceEvent` (`src/messages.ts:298`), `isAgentInstructionsRow` (`:291`), `newestInstructionSeqsOf` (`src/region.ts:880`), `guardedSurfaceSeqsOf` (`:922`), the barrier at `src/region.ts:1012`, and the hard reject in `src/tools.ts:416` / `:454`.

1. **For the loop #93 fixed, B2 has nothing to elide.** The host's gate is identity-based, not content-based (`origin/main:src/region.ts:865-866` — a file's instructions are re-injected when its *current copy is absent* from the surface). #93 keeps the current copy on the surface, so AGENTS.md is simply never re-injected. Your own history comment records the measured value as zero (`src/injection-dedupe.ts:13-15`).
2. **The stub can steal #93's newest-row pin, and that path loses content.** `src/injection-dedupe.ts:132` returns `{ ...message, content: [{ type: 'text', text: stub }] }` — the spread keeps `source`, including `source.changes[].scope`. The *stub* therefore becomes the newest `agent-instructions` row of that scope: `newestInstructionSeqsOf` (`origin/main:src/region.ts:901`) maps the scope to it and `guardedSurfaceSeqsOf` (`:928`) guards it, while the real full row becomes a *stale copy* — and main deliberately lets stale copies be compressed (`origin/main:src/tools.ts:446`). Once that compress lands, the only AGENTS.md bytes are gone and the host will not re-inject, because the presence gate is identity-based and the stub carries the identity. That is exactly the loss your fix note claims to prevent. The stub is also classified `'instruction'`, so it is a permanent barrier (`origin/main:src/region.ts:1012`) and, as newest, un-compressible — stubs accumulate as un-removable surface nodes.
3. **One test encodes behavior main now forbids.** `tests/injection-governance.test.ts:148-156` compresses an instruction row to prove elision — the very operation `src/tools.ts:454` hard-rejects at the tools layer. The unit seam (`runCompactionTransaction`) does not guard, so the test passes while asserting a forbidden invariant.

## The three conflicting files

- **`src/index.ts`** — keep main's 5-arg `buildNudge` call with the emergency-budget callback (`origin/main:src/index.ts:321`, `:325-336`); wrap only the messages (dedupe `decision.messages`, then append `outcome.message`). Dropping those arguments silently deletes the #108 per-turn budget. Keep your `!autoNudge` dedupe branch.
- **`src/prompts.ts`** — adjacency only: keep both main's paged-`decompress` copy (`origin/main:src/prompts.ts:210`) and your `verifiedReadings` sentence.
- **`src/region.ts`** — imports: union. `AcpCompactionSummaryFields`: main deleted the six-field interface in favor of `readCompactionSummary` (`origin/main:src/region.ts:388`) — don't resurrect the top-level fields (see P0-3).
- **`src/messages.ts`** (auto-merges, but needs a decision) — you now carry two checkpoint predicates: main's `isCheckpointNode` (`origin/main:src/messages.ts:222`) and your `isCompactionCheckpoint` (`src/messages.ts:133`). Keep one.

## Smaller items

- **P1** the whitelist misses the second host shape: `src/injection-dedupe.ts:30` covers `kind: 'agent-instructions'`, but `origin/main:src/messages.ts:294-296` also accepts `kind: 'plugin'` + `plugin: 'agent-instructions'`. If the live hook emits the latter, your headline case is never elided.
- **P1** duplicates inside one batch are not elided: `visible` is computed once (`src/index.ts:310-311`) and the frames kept earlier in the same call are never added to it (`src/injection-dedupe.ts:115-133`).
- **P1** the stub drops non-text content — `src/injection-dedupe.ts:132` replaces `content` wholesale, so any image/tool block riding `runtime` / `runtime-context` is discarded (compare main's `REAL_CONTENT_PLUGINS` caution for `tools-ptc`).
- **P1** B3 is write-only: `verifiedReadingsOf` has no production caller (only `dist/*`, the definition at `src/region.ts:393`, and tests). Nothing surfaces the readings to the model, so "later steps need not re-run the checks" is unreachable as shipped.
- **P2** `src/region.ts:447` stores the raw `summary` while `:468` stores the prefixed one — the durable event and the durable node disagree, so a reader using the event gets different text than context/search shows.
- **P2** `SUMMARY_FRAME_PREFIX` is Chinese-only inside an otherwise English prompt surface, embeds a directive into every summary, and adds CJK tokens to every checkpoint's price.
- **P2** `visibleInjectionKeys` re-reads and SHA-1s the full text of every visible injection row on every `agent/pre-step`, with no memoization.

## Tests

- Real fixtures: only `appendInjection` (`tests/injection-governance.test.ts:31-36`) matches a real host shape. `msg()` (`:27`) is never validated against the actual injected frame, and the "tool" case at `:111` uses `source: { kind: 'tool' }` rather than the real nested `tool-result` shape.
- Nothing drives `agent/pre-step`, so the P0-1 no-op is invisible to the suite.
- B6-2 / B6-3 (`tests/g5-governance.test.ts:69-93`) are tautological: the fixture is built from the same imported kernel constants, so a kernel copy change passes while production stripping silently stops matching.
- B3-1 asserts `readCompactionSummary(event).verifiedReadings` — after the P0-3 fix that line must change or the assertion goes vacuous.
- B2-5 (`:139-162`) encodes the pre-#93 invariant.

## Ranked fixes

1. Rebase onto `origin/main` and use `sessionEventsOf` / `eventAtOf` (`src/injection-dedupe.ts:81`); fix the tests (`tests/injection-governance.test.ts:56,73,84`).
2. Move `verifiedReadings` into the block ledger (`src/block-ledger.ts:37-85`), delete the top-level write, keep `rawOutput`.
3. Resolve the stub ↔ #93 identity conflict — either strip `source.changes[].scope` / `kind` from the stub so it can never win `newestInstructionSeqsOf` (and document why), or drop B2.
4. Dedupe within a single batch.
5. Delete `isCompactionCheckpoint` and reuse `isCheckpointNode`.
6. Make B3 readable (compress result and/or `acp_status`) or drop it.
7. Tests: real tool-result fixture; an integration/e2e test that drives `agent/pre-step` and asserts elision in the *request* messages; replace the tautological B6 tests with a pinned expected substring.
8. Reconcile the prefix/raw mismatch, and consider an English marker consistent with the rest of the tool copy.
9. Drop the `dist/` commits — current policy is that contributors don't commit build output and a bot rebuilds it after merge.

## Questions I can't answer from the branch

1. Which host channel re-injects a frame **while an identical copy is still visible** on the surface, post-#93? If none, B2's byte savings are zero and I'd drop it.
2. Does the pre-step injected frame carry `source.changes[].scope`? If yes, the stub steals #93's newest-row pin (the loss path above). If no, the stub is an un-guarded, un-removable node.
3. Which shape does AGENTS.md actually arrive as — `kind: 'agent-instructions'`, or `kind: 'plugin'` + `plugin: 'agent-instructions'`?
4. Can `runtime` / `runtime-context` frames carry image or non-text blocks?

Happy to take any of this over if you'd rather not rebase — just say which parts you want to keep. B1 and B6 look ready to land on their own.

ranxianglei:

Review taken over — all three blockers were re-verified against `origin/main` (`baa8834`) before any code changed, then the ranked fixes were implemented. Summary:

**Pre-fix verification (all confirmed)**
- **P0-1**: `src/injection-dedupe.ts:81` read `session?.events ?? []`; dsh-session 0.1.5 exposes no `events` getter (only `eventAt()`/`snapshotEvents()`), so `visibleInjectionKeys` silently returned an empty set and B2 was a guaranteed no-op.
- **P0-2**: the PR's own tests indexed `session.events[…]` and crashed at runtime under `npm test` on a real `Session`.
- **P0-3**: top-level ledger members vs main's rawOutput encoding (`encodeAcpBlockLedger`, `src/region.ts:447`) — the frozen v0 reader's exact member allow-list rejects unknown members (issue #141 class).

**Your four questions** (audited against the published `@deepseek-ai/*` 0.1.5 packages):
1. **No channel re-injects while an identical copy stays visible.** dsh-agent-loop's `RuntimeContextProjection` skips when the retained snapshot text is identical (`if (this.retained?.text === snapshot) return;`); agent-instructions re-inject only once the current copy disappears from the surface (rule 16 live audit; the author's own audit: 0 bytes saved, 2 content losses). Per your criterion → **B2 dropped**.
2. **Yes.** The frozen v0 reader validates exactly `{kind:'agent-instructions', form:'instructions', changes[]}` with strict keys `{action: set|replace|remove, scope, path, digest?}` — so a spread-stub would have won the newest-row pin via `newestInstructionSeqsOf` while the FULL copy degraded to a compressible stale row: content loss with no re-injection.
3. Current shape is `kind:'agent-instructions'` (the legacy `kind:'plugin' + plugin:'agent-instructions'` still passes validation). Real runtime-context frames carry `kind:'plugin'` + `plugin:'@deepseek-ai/dsh-system-prompt'` — so `DEDUPE_SOURCES` matched at most one real frame kind regardless.
4. Runtime-context frames are always a single text block (the projection builds `[{type:'text', text: snapshot}]`); non-text content is not a real shape there.

**What landed** — new branch `ranxianglei:2026-09-11_injection-governance-b1-b3-b6`: the original commit preserved verbatim in history, plus one review-fix commit on top.
- Rebasing onto `baa8834`; conflicts resolved in `src/index.ts` / `src/prompts.ts` / `src/region.ts` (kept main's memoized ledger, paged decompress description, 5-arg `buildNudge`).
- **B2 dropped entirely** (`src/injection-dedupe.ts` deleted, pre-step wiring back to main shape).
- **B1**: English marker `[Model-written summary — not user words; re-verify any obligations before relying on them]`; private `isCompactionCheckpoint` deleted in favor of main's `isCheckpointNode`; one `framedSummary` written identically into the `compaction/summary` event AND its checkpoint node; projection-time framing kept as an idempotent legacy-only net.
- **B3**: `verifiedReadings` moved into the rawOutput block ledger (seventh field, omitted when empty) — never top-level; read face `verifiedReadingsOf(event)` decodes embedded ?? legacy top-level, `Array.isArray`-guarded, never throws; surfaced on `AcpBlockLedgerEntry`; echoed in the compress result (`, verified: <a; b>`, non-empty only).
- **B6**: kept; the tautological constant round-trip test replaced with pinned substrings of REAL rendered nudge output on both the kernel and template paths (measured body ≈224 B ≤ 300).
- Tests rewritten on the real seam API (`sessionEventsOf`/`eventAtOf`); the M3 decompress test now expects the framed summary header.
- Docs synced in the same change: new `docs/injection-governance-design.md` (decisions + host evidence), README.md / README.en.md feature notes, AGENTS.md module map + new hard-won rule 17, docs/README.md index.
- `dist/` intentionally out of the diff (dist-bot refreshes after merge).

**Verification**: `npm run typecheck` clean · `npm test` 273/273 · `npm run build` clean · `npm run test:e2e` all scenarios pass.

**One logistics note**: this bot's account has fork-only access — it cannot push to this repo or edit #132's head ref (both attempts 403/404), so the final state is carried by a superseding cross-repo PR: **#143** — https://github.com/Tyan66666/billion-context-dsh/pull/143 . Please close #132 as superseded (or retarget it to the same branch if you'd rather keep the number). @TheTouYu — your design and commit stay credited as-is in the history; if a future host genuinely re-injects over visible copies, B2 can come back redesigned against the real shapes (the design doc spells out the conditions).

修复了 review 指出的三个阻塞项（B2 在真实 seam 上是静默空操作、测试对真实 Session 崩溃、顶层账本字段会重新砖化升级前日志）并按排名方案落地：整体移除 B2（宿主不存在“副本仍可见时重复注入”的通道，存根反而会抢占 #93 最新行保护造成内容丢失）、B1 摘要标源改为英文标记并在创建时一次性同写事件与检查点节点、B3 verifiedReadings 移入 rawOutput 账本且在 compress 结果回显使其可读、B6 保留并把测试改为锚定真实渲染输出；已 rebase 至 main，typecheck / 273 项单测 / 构建 / e2e 全部通过，可合并 #143。

Tyan66666:
Closing as **superseded by #143**, which is now merged into `main`.

## What happened

My review above raised three P0 blockers. @ranxianglei re-verified all three against `origin/main` and implemented the ranked fixes on a fresh branch, shipped as cross-repo PR #143 (the bot has fork-only access, so it could not push to this PR's head ref — that head branch has since been deleted here and this PR is left DIRTY with no CI). #143 merged on 2026-09-11T14:01:04Z.

## What landed via #143

- **B1 — summary source framing.** `SUMMARY_FRAME_PREFIX` / `withSummaryFramePrefix` / `prefixSummaryBlocks` (`src/messages.ts`, `src/region.ts`), with the English marker `[Model-written summary — not user words; re-verify any obligations before relying on them]`. The frame is written once at creation, identically into the `compaction/summary` event AND its checkpoint node (the old code stored the raw summary in one and the prefixed one in the other), and the projection-time framing stays as an idempotent net for legacy blocks only. Your two checkpoint predicates collapsed into one — main's `isCheckpointNode`.
- **B3 — verified readings.** Moved into the `rawOutput` block ledger (`encodeAcpBlockLedger` / `decodeAcpBlockLedger`, seventh field, omitted when empty), never a top-level `compaction/summary` member — so the frozen v0 reader's member allow-list is satisfied and pre-upgrade logs stay openable (this was P0-3: the original form re-broke the issue #141 class). Read face `verifiedReadingsOf(event)` decodes embedded ?? legacy top-level, `Array.isArray`-guarded, never throws; it is surfaced on `AcpBlockLedgerEntry` **and echoed in the compress result**, which was the missing piece that made B3 write-only.
- **B6 — slim nudges.** `stripNudgeGuidance` on both the kernel-default and template paths; the tautological constant round-trip test replaced with pinned substrings of real rendered nudge output.

Documentation shipped with it: `docs/injection-governance-design.md`, README/README.en feature rows, and AGENTS.md module map + a new hard-won rule.

## What was dropped

- **B2 — unchanged-injection elision.** No host channel re-injects a frame while an identical copy is still visible: dsh-agent-loop's `RuntimeContextProjection` skips on an identical retained snapshot, and `agent-instructions` re-inject only once the current copy disappears from the surface — so for the loop #93 fixed there was nothing to elide, and the measured value was zero. Worse, the spread-stub kept `source.changes[].scope`, which would have won `newestInstructionSeqsOf` while the full copy degraded to a compressible stale row: content loss with no re-injection. The design doc records the conditions under which B2 can come back redesigned against real shapes.
- **B4/B5 — the state face** (`revokeIf` plus the 300-byte `FACE_BUDGET`). Neither landed nor tracked anywhere; if it is still wanted, it should be a fresh issue with its own evidence rather than riding this PR.

## Credit

@TheTouYu — your design and commit stay credited in #143's history, and the mechanism is credited in the docs from that PR (`docs/injection-governance-design.md`, AGENTS.md). B1 and B6 were the genuinely useful half of this change set, and they are in `main` because you wrote them.

PR #123, opened 2026-09-06
perf(region): memoize rebuildBlockLedger per snapshot to cut O(B^2*N) rebuilds (#109)

ranxianglei:
## What
Fixes #109 — `rebuildBlockLedger` was being recomputed many times inside hot loops instead of once per tool call.

## Root cause
`rebuildBlockLedger` scans the full durable log **and** calls `summarySeqOfCompaction` (itself O(N)) once per compaction block → **O(B·N)** per call. Several helpers wrap it (`blockRegistry`, `expandShadowedSeqs`, `blockRefForSummarySeq`, `compactionIdsOfKernelBlocks`, `summarySeqOfKernelBlock`) and are invoked **per element** in loops:

- `search_context` → `buildSearchDocs` loops `expandShadowedSeqs` over every block → **O(B²·N)**
- `compress` → `handleCompress` calls `blockRefForSummarySeq`×2 + `compactionIdsOfKernelBlocks` per range → **O(R·B·N)**
- tier-nudge → `summarySeqOfKernelBlock` per target block

So search / compress / nudge latency grew super-linearly with the number of blocks.

## Fix
Memoize `rebuildBlockLedger` on the input snapshot array via a module-level `WeakMap`, guarded by array length. Safe because:

- within one tool invocation every caller passes the same stable snapshot (`sessionEventsOf` reuses it until the next append);
- the log is append-only, so same reference + same length ⇒ identical content;
- after an append the snapshot is either a new array (copy-on-write) or grows in length → the key misses → recompute.

The function is pure and no caller mutates the returned ledger, so sharing the cached array is correct. Net effect: each tool call rebuilds the ledger **once** (O(B·N)) instead of per element — a ~B× reduction on the hot paths. Zero signature changes; all callers benefit automatically.

## Verification
- `npm run typecheck` ✅
- `npm test` → **209/209** pass (added an idempotency regression test asserting repeated calls on one snapshot return an identical ledger)
- `npm run build` ✅ — `dist/` regenerated for CI's dist-drift check; `dist/region.d.ts` is byte-identical to main (no public type change)

## Files
- `src/region.ts` (+7): the `WeakMap` memo + length guard
- `tests/region.test.ts` (+21): idempotency regression test
- `dist/index.js`, `dist/index.js.map`: rebuilt bundle

Issue #109, opened 2026-09-06
[perf] rebuildBlockLedger/blockRegistry 未缓存，search/compress/nudge 内被循环 O(N) 重扫


ranxianglei:
来源: ranxianglei/billion-context-pi#305 分析 "对照本项目分析billion-context-dsh中存在的bug" 时发现

## 现象
`rebuildBlockLedger`（O(N) 扫全日志）与 `blockRegistry` 每次调用都重建、无 memoization/LRU，却被多处路径在循环里反复调用，导致单次操作成本随块数/范围数放大。

## 证据
- `src/region.ts:477-516` `rebuildBlockLedger` O(N)、无 memo
- search_context ≈ O(块数 × 日志)：`src/tools.ts:601-636` `buildSearchDocs` 在 `for (const block of ledger)`（`:605`）内每块调 `expandShadowedSeqs`（`:615`），后者又 `rebuildBlockLedger`（`region.ts:1051`）
- compress ≈ O(范围数 × 日志)：`src/tools.ts:381-382` 每 range edge 调 `blockRefForSummarySeq`（→`region.ts:1001`→`blockRegistry`）；`:483` 每 range 调 `compactionIdsOfKernelBlocks`（→`region.ts:1009`）；`:481` 每 range 调 `shadowedTokensViaMeter`（内部 `meter.measure` 全会话，`host-tokens.ts:137`）
- tier nudge ≈ O(target × 日志)：`src/nudge.ts:245/319` `summarySeqOfKernelBlock`（→`region.ts:1031`→`blockRegistry`）每 target 一次
- decompress：`blockIdOfKernelRef`/`resolveBlockId`/`expandShadowedSeqs` 多次全量重建

## 影响
长会话下 search_context / compress / status 显著变慢；随会话总长线性甚至准二次增长。

## 建议修复
单次操作内只算一次 `rebuildBlockLedger`/`blockRegistry`，把结果 thread 进各 helper（给 `blockRefForSummarySeq`/`compactionIdsOfKernelBlocks`/`summarySeqOfKernelBlock`/`expandShadowedSeqs` 增加可选 ledger 入参）。

ranxianglei:
我来协助看一下这个问题,请分析并回复处理结果。

Tyan66666:
Closing: PR #122 landed the caching this issue asked for. Verified on main:

**Fixed**
[...]

PR #134, opened 2026-09-09
(fix) search_context: cache the block ledger and search corpus per log snapshot — 20-minute searches become sub-second

### LINKED ISSUE #133: search_context 随压缩块累积而急剧变慢（从毫秒级劣化到 ~20 分钟）——疑似每调用 O(B²·N) 的账本重建 + 全量语料重复分词
## 摘要

`search_context` 的时延随会话内压缩块数量近似单调恶化。在我们的一个长会话中，当压缩块达到约 190 个、被遮蔽 token 约 1300 万时，单次 `search_context` 耗时约 **20.8 分钟**。我认为这里面存在可避免的算法级低效：每次调用都会**从零重建**整个块账本与搜索语料，并且混合打分器会对**全部被遮蔽原文**重复分词。

## 运行环境

- `billion-context-dsh` **v0.2.19**（自包含 `dist/index.js`）
- 宿主：DeepSeek Harness web 端（`@deepseek-ai/dsh-web-app`，平台 `dsh@0.1.2-rc.1`）
- peers（profile 实际解析）：`@deepseek-ai/cordis@4.0.2`、`@deepseek-ai/dsh-compaction@0.1.2-rc.1`、`@deepseek-ai/dsh-session@0.1.2-rc.1`
- 均使用默认配置（无 `coreOverrides`、无 `countTokens` 覆盖）

## 实测劣化

`tool/call` 与 `tool/result` 事件都带 epoch 毫秒 `time`，`result.time − call.time` 即工具真实耗时；本次 8 次调用全部成功配对。

| 调用 | 当时块数 | 当时遮蔽 token | 实测耗时 |
|---|---|---|---|
| 1 | 6 | ~0.7M | **1.8 s** |
| 2 | 6 | ~0.7M | **0.4 s** |
| 3 | 55 | ~4.0M | **94 s** |
| 4 | 141 | ~10.1M | **373 s（6.2 分钟）** |
| 5 | 141 | ~10.1M | **418 s（7.0 分钟）** |
| 6 | 158 | ~11.2M | **639 s（10.7 分钟）** |
| 7 | 158 | ~11.2M | **702 s（11.7 分钟）** |
| 8 | **190** | **~13.3M** | **1249 s（≈20.8 分钟）** |

会话结束时累计：246 个块、17.6M 遮蔽 token、1.15M 摘要字符、未压缩日志约 274MB。每次搜索要把 **约 1.94 万条被遮蔽消息**展开进文档集（其中约 43% 是 `tool/result`），即约 1.97 万个文档、几十 MB 文本，且整体重新分词。

时延随「块数 × 遮蔽 token」近似线性上升——会话压缩得越深，后续每次搜索就越慢，形成正反馈恶化。

### A. 每次搜索 O(B²·N) 的块账本重建（插件层，主导耗时）

`buildSearchDocs(session)`：
- 先调用一次 `rebuildBlockLedger(sessionEventsOf(session))`，
- 然后**对每个块**调用 `expandShadowedSeqs(session, block.blockId)`。

`expandShadowedSeqs(...)`：
- **每次调用都从零重跑一遍 `rebuildBlockLedger(sessionEventsOf(session))`**。

`rebuildBlockLedger(events)`：
- 遍历全部事件一次（O(N)），
- 并且对**每个** `compaction/summary` 条目调用 `summarySeqOfCompaction(events, compactionId)`——后者**线性扫描整个事件列表**寻找对应压缩检查点（O(N)）。

合计：单次 `search_context` ≈ `(B+1) × B × N` 次数组遍历。以 B=246、N≈16 万计，约 **97 亿次迭代**——这还没算任何分词。账本对给定会话是不可变的（只有新块落盘才变化），完全应该只算一次并缓存；`summarySeqOfCompaction` 也可以用单遍 `Map<compactionId, seq>` 或从 summary 事件自身 seq 起的有界前向扫描替代。

### B. 搜索语料 = 每个块的摘要 + 全部被遮蔽原文，每次调用原样重建（插件层）

`buildSearchDocs` 为每个块（摘要文本）以及每一条被递归展开的被遮蔽消息各建立一个文档；并没有跨调用的增量/倒排索引——每次搜索都重新构造全量约 2 万文档的语料。

### C. 特征缓存上限导致复用失效（内核层，acp-kernel）

`docFeatures` 以「整段文本」为 key 缓存分词/二连字特征，但上限为 `DEFAULT_CAP_CHARS = 8 * 1024 * 1024`（字符 8MB）且 LRU 逐出。语料有几十 MB，远超上限，因此**每次调用都要对超出部分重新分词**（每个文档 `Intl.Segmenter("zh")` 词切分 + 逐文档字符二连字 `Set`）。

### D. 混合打分对全量语料双跑

`hybridAlgorithm` 同时跑 BM25（词干 + CJK 分词，扫全部文档）与 fuzzy（二连字重合，扫全部文档），随后 `applyRoleWeight` 与 `buildResults` 又各对全量文档建一次 `Map`。对小型语料没问题，在 1.97 万文档 / 几十 MB 时就是灾难。

## 修复建议（按收益排序）

1. **按会话缓存块账本**（仅在新增 `compaction/summary` 时失效）——立竿见影去掉 `(B+1)` 倍放大。（插件层）
2. **单遍建 `compactionId → 检查点 seq` 索引**（或从每个 summary 事件自身 seq 处前向扫描），让 `rebuildBlockLedger` 从 O(B·N) 降到 O(N+B)。（插件层）
3. **持久/增量维护搜索索引**（文档随块落盘只会追加），避免每次全量重建+全量分词；上调或重构 8MB 特征缓存上限（改为按文档 id/块做 key，而非整段文本）。（内核层）
4. 可选：默认**先只搜摘要**，把「搜被遮蔽原文」做成显式选项（绝大多数查询摘要即可满足，需逐字原文时用 `decompress` 兜底）。（插件层）

如果需要，我可以直接为第 1~3 项提 PR——改动面小、边界清晰。

## 备注

- `decompress` 与分层蒸馏路径同样受影响（共用了 `expandShadowedSeqs`/`rebuildBlockLedger`）。
- 本 issue 中的数值全部取自压缩块元数据（块数、遮蔽 token 数、消息数）的聚合量，不含任何会话内容。

ranxianglei:
Fixes #133. Related: #109 (same root cause, plugin layer); this PR supersedes #123 (see "Ledger memoized" below).

## Problem

`search_context` latency degrades monotonically as compressed blocks accumulate: in the live session reported in #133, a single search took 1.8 s at 6 blocks and 1249 s (~20.8 min) at 190 blocks / 13.3M shadowed tokens. The deeper a session is compressed, the slower every subsequent search becomes.

## Cause

Two plugin-layer hot spots (verified in the v0.2.19 code, reproduced locally at the reported scale):

1. **The block ledger was rebuilt (B+1) times per search, and each rebuild was O(B·N).** `buildSearchDocs` called `rebuildBlockLedger` once, then `expandShadowedSeqs` per block — each of which re-ran the full ledger from scratch. And the ledger itself scanned the whole event log linearly per block (`summarySeqOfCompaction`) to find each block's checkpoint seq. Total ≈ (B+1)×B×N array iterations: ~5.8B at B=190 / N=160K.
2. **The search corpus (block summaries + ALL shadowed originals — ~20K docs / 40–50MB in the reported session) was re-extracted and re-tokenized on every call.** The kernel's `docFeatures` cache (8MB cap, insertion-order eviction) is far smaller than the corpus, so nothing was reused across calls: measured 17.9 s per search, identical cold and warm.

## Fix

- **Single-pass ledger** (src/region.ts): one scan of the event log builds a `compactionId → checkpoint seq` index (`summarySeqIndex`); `rebuildBlockLedger` goes from O(B·N) to O(N+B). Semantics unchanged — first checkpoint per compactionId wins, same as the old per-block scan.
- **Ledger memoized per log snapshot** (src/region.ts): the ledger is cached in a `WeakMap` keyed on the append-only snapshot array (identity + length check). The host session log returns the same frozen array until the next append (verified on both seam shapes: 0.1.0-rc.6 `get events()` and 0.1.2 `snapshotEvents()`), so repeated reads on an unchanged log are O(1). This is the same approach as #123 (issue #109) — included here because the single-pass fix alone still left (B+1) rebuilds per search (~400 ms at the reported scale); the two fixes are complementary, and merging this PR makes #123 redundant.
- **Search corpus cached per log snapshot** (src/tools.ts): `buildSearchDocs` returns the cached `SearchDoc[]` until the next append; a new block invalidates it automatically (same snapshot-array keying).
- **Kernel doc-cache cap raised to 128MB** (src/index.ts): the engine constructor calls the kernel's public `setDocCacheCap` tuning API. The 8MB default re-tokenized the whole corpus on every search; 128MB covers the largest reported session (17.6M shadowed tokens ≈ 70MB text). Steady-state feature memory is ~4–6.3× cached source chars (measured), so worst case ≈ 800MB heap — acceptable for a host that already holds logs of that scale.

## Verification

Reproduced the reported scale locally (190 blocks / 40.4MB corpus / 20,140 docs) and measured before/after:

| phase | before | after |
|---|---|---|
| ledger rebuild ×1 | 39 ms | 2 ms |
| corpus build (`buildSearchDocs`) | 7,227 ms | 375 ms |
| search (first, cold tokenization) | 17,933 ms | 8,992 ms (one-time) |
| search (subsequent, warm) | 17,436–17,954 ms | 140–161 ms |

At the max scale tested (160K events / 160,170 docs): corpus build 101,367 ms → 584 ms. Repeated searches before the next append rebuild nothing; the positive-feedback degradation (more blocks → slower searches) is gone.

- New regression tests (tests/search-perf.test.ts, 4 tests): ledger memoization + invalidation on append, first-checkpoint-wins index semantics, corpus cache identity + invalidation, and a perf guard (10K events / 150 blocks corpus build < 500 ms — the old code takes seconds at this scale; verified the new suite fails on the pre-fix code).
- Full suite: 218/218 pass (was 214); `tsc --noEmit` clean. Docs updated in the same PR (README tool table, AGENTS.md rule 13).

## Out of scope (kernel layer, filed upstream)

- The 8MB `docFeatures` cap design (a multi-session tradeoff) — a design call for the kernel owner: ranxianglei/acp-kernel#227.
- **New bug found while reproducing at scale**: `hybridAlgorithm.score` normalizes with `Math.max(...scores)` — an argument spread that throws `RangeError: Maximum call stack size exceeded` at ~125K+ docs (V8 argument limit; reproduced at 160K docs). Search hard-crashes instead of degrading for sessions with ~30K+ compressed messages. Fixed upstream in ranxianglei/acp-kernel#228 (loop-based max + 130K-doc regression test); bump the pin when it lands.
- Suggested fix #4 in #133 (summaries-only default search) is a behavior change — left for the owner to decide.


Tyan66666:
Pushed two commits to the branch so the PR is mergeable as-is (`56ce143` = merge of main, `21faecc` = review follow-ups). Four things needed fixing:

**1. The tests were red on the current seam.** `tests/search-perf.test.ts` read `session.events`, a getter the host removed in 0.1.5. Re-running the pre-fix file against real `Session` objects gives `# tests 4 / pass 2 / fail 2` — `TypeError: Cannot read properties of undefined (reading '1')` from `buildPerfSession`. They now read the snapshot through `sessionEventsOf(session)` (`src/session-events.ts`), the same accessor the engine itself uses. `npm test` is 232/232 after the change, and both tests still fail on the pre-fix code, so they remain a real regression gate.

**2. The cap arithmetic was off by one to two orders of magnitude.** The kernel's cap bills *source* chars, but what an entry retains is its feature heap. Measured on 2M-char corpora (`--expose-gc`, full GC, the corpus itself excluded so it cannot inflate the delta):

| content | retained heap / billed char |
|---|---|
| latin prose | 1.9× |
| CJK prose | 9.3× |
| high-entropy latin (code / logs / JSON) | 26.4× |
| high-entropy CJK | 48.9× |

So `setDocCacheCap(128MB)` costs ~270MB–6.5GB rather than the "~800MB worst case" — and the 70MB session the PR cites lands at ~0.7–1.7GB. The multiplier tracks lexical diversity, not text length, which is why the low-diversity estimate did not hold up.

**The cap also cannot be lowered to compensate.** The kernel evicts FIFO and bills source chars only, so any cap below the corpus caches nothing at all (measured on a 2M-char corpus, repeat full scan: cap = half the corpus → 1.1×; cap = full → 1177×–9522×). Given that step function 128MB stays — it is the only setting where the caching works for the sessions this PR targets — but the trade-off is now stated instead of understated: `AGENTS.md` rule 14 carries the numbers, and the constructor comment points at it.

**3. Rule number collision.** The new rule was numbered 13, which main already assigns to the session-replace-protocol rule (#136). It is now rule 14.

**4. `dist/` had not been rebuilt for the src changes** — `dist/index.js`, `dist/index.js.map` and `dist/tools.d.ts` were stale on the branch, so the committed bundle did not match the source. Refreshed. (CI's dist check only runs on schedule / from the dist-bot, so this would not have failed the PR, but §5 of AGENTS.md asks for it in the PR itself.)

One design question worth settling before this merges: `setDocCacheCap` writes a module-level value in the kernel, so the setting — and the memory it authorises — is process-wide, shared by every acp-kernel consumer in the process. In a DSH profile this plugin is the only consumer, so it is fine today; an instance- or session-scoped option upstream would remove the coupling if that ever changes.

Verified locally: `npm run typecheck` exit 0; `npm test` 232/232; `npm run build` clean. The perf test's only wall-clock gate (`firstMs < 500`) runs at 9.7ms measured — a 51× margin, so it should not be flaky in CI.

ranxianglei:

Reviewed on head `21faecc`: `typecheck` exit 0, `npm test` **232/232**, `npm run build` clean. All four follow-ups check out, and I settled the design question below. One thing to pull out before merge (scope) — see the last section.

**1. Tests red on the current seam — confirmed.** `tests/search-perf.test.ts` now reads the snapshot through `sessionEventsOf(session)` (import at line 17; used at line 32 and inside `buildPerfSession` at line 104) — the same accessor the engine uses. It drives real `Session` objects (`Session.create('perf-133')` + `createUserMessage`/`createAssistantMessage`), so the old `session.events` read (getter removed in 0.1.5) is gone. The four guards are meaningful: ledger memoization + invalidation-on-append, first-checkpoint-wins index semantics, corpus identity + invalidation, and the wall-clock gate `assert.ok(firstMs < 500)` (line 137) over a 10K-event / 150-block corpus. Real regression gate, not a vacuous one.

**2. Cap arithmetic — agreed, and it now lives in rule 14.** The kernel bills *source* chars but retains a *feature heap*; your measured 1.9×–48.9× multiplier (lexical-diversity-driven, not length) supersedes the "~800MB worst case." 128MB ⇒ ~270MB–6.5GB, ≈0.7–1.7GB for the cited 70MB corpus. And because eviction is FIFO and bills source chars only, any sub-corpus cap caches nothing (your half-corpus → 1.1× result), so 128MB is effectively the floor that works. Rule 14 states the trade-off instead of understating it. Good.

**3. Rule number — fixed.** The new rule is 14; rule 13 stays the session-replace-protocol rule (#136). No collision remains in the net diff.

**4. `dist/` refresh — substantively done, one cosmetic caveat.** The committed bundle **does** carry all the new logic — marker counts identical between committed and a fresh build (`setDocCacheCap` 2/2, `summarySeqIndex` 2/2, `blockLedgerCache` 3/3, `searchDocsCache` 3/3). A fresh `npm run build` here differs only by a build-environment comment path: `// ../../node_modules/acp-kernel/dist/index.js` vs `// node_modules/acp-kernel/dist/index.js` — an exact 6-byte delta that cascades into the sourcemap offsets. That's an esbuild/tsup artifact of checkout directory depth, not code drift. It won't block the PR (the dist check no longer gates PRs) and the dist-bot regenerates `dist/` after merge. Non-blocking — noting it only so a spurious drift flag later isn't a surprise. If you ever want byte-stable `dist/`, normalize that banner path.

**Design question — `setDocCacheCap` process-wide:** you're right, and it's broader than the cap number. I checked the kernel source: `setDocCacheCap` mutates a module-level `capChars`, but the feature cache itself is *also* a single module global (`var cache = new Map()`, `var cachedChars = 0`) shared by every acp-kernel consumer in the process. So an instance-/session-scoped option upstream would have to scope both the cap and the cache, not just the number. My recommendation: **accept as-is for this PR.** In a DSH profile this plugin is the only acp-kernel consumer, so there is nothing to collide with today. Don't open a fresh upstream issue — the cap-design question is already tracked on acp-kernel#227; add the "the cache Map is module-global too, so scoping must cover cap + cache" note to that thread when you next touch it. Keep the constructor comment pointing at rule 14. That records the coupling without stalling the fix.

**Scope — the thing to settle before merge:** commit `a6e55a5` bundles an unrelated feature, **#140** (nudge context breakdown re-pointed to the live surface: `computeSurfaceBreakdown` + `isCheckpointEvent` in `src/nudge.ts`, `tests/nudge.test.ts`, the rule-2 expansion, the verification-doc row, `nudge.d.ts`). It's a good, self-contained, display-only change — but it has nothing to do with `search_context` performance, and the repo's first-gate rule is that a dirty diff (unrelated changes riding along) isn't merged. A little chore noise rides along too (`.gitignore` local artifacts, markdownlint table-format churn in `docs/dsh-porting-verification.md`).

Recommendation: **split #140 (and the chore noise) into their own PR(s), leaving #134 as the pure search-perf fix.** Everything here is green and correct, so the split is mechanical — branch off `main` at the merge point and cherry-pick just `a346da2` (search-perf) plus the relevant parts of `21faecc`, moving `a6e55a5` onto a fresh #140 branch. Happy to help prepare the split if useful. If you'd rather keep them together that's your call as owner — but on the stated standard I'd hold the merge until #134's diff is only the search fix.

**Bottom line:** the search-perf fix itself is correct and fully verified; the only thing between this and merge is pulling the unrelated #140 out.

一句话中文总结:搜索性能修复本身正确且已验证全绿(typecheck / 232 项测试 / build 均通过,回归测试真实有效,cap 内存测算已如实写入规则 14),但 PR 里混入了与本 PR 无关的 #140 nudge 展示改动和少量 chore,按仓库"脏 diff 不合并"规范建议先把 #140 拆成独立 PR,#134 只保留搜索缓存修复后即可合并;`setDocCacheCap` 进程级共享(含特性缓存本身就是模块级单例)今天在本插件独占场景下无冲突、可接受,后续把该观察并入上游 acp-kernel#227 讨论即可。

PR #138, opened 2026-09-10
(fix) compress emits the current SurfaceOp protocol — startSeq/endSeq, not start/end (issue #136)

hairyf:
## Problem

Every `compress` call fails on current DSH Desktop (billion-context-dsh@0.2.21 with `@deepseek-ai/dsh-session@0.1.5-alpha.2`) with:

```
compress failed: session event "user/message" carries an invalid replace surfaceOp
```

## Cause

`dsh-session >= 0.1.3-alpha.2` validates replace `surfaceOp`s as **exactly** `{ op, startSeq, endSeq }` (`Object.keys(op).length === 3 && hasOwn(op,'startSeq') && hasOwn(op,'endSeq')`), but the engine emitted the old field names `{ op, start, end }`. The breaking change is deepseek-harness commit [27bf1039db](https://github.com/deepseek-ai/deepseek-harness/commit/27bf1039db) "refactor(session)!: distinguish event seqs from log offsets" (2026-08-30), which renamed the fields. The package's peer range stopped at `^0.1.2-alpha.4`, so the newer session installed anyway and compress broke at runtime.

A second, latent incompatibility: the same protocol change makes an **empty `assistant/message` replace impossible** — `sourceEventSeqs` are forbidden on `assistant/message`, so a replace that shadows a non-empty range can never cite its sources and is rejected on replay. The engine's orphan-cleanup / empty-result hide used exactly that shape and would crash at append on new hosts.

## Fix

1. **Emit the new field names** in all three replace sites of `src/region.ts` (`runCompactionTransaction`, `hideSurfaceSeqs`): `surfaceOp: { op: 'replace', startSeq, endSeq }`.
2. **Convert the empty-assistant prune hide to a terse `user/message`** (the only replace idiom the new protocol accepts for a shadowed range — same visible-node idiom as the host's compaction-basic and tool-result pruner).
3. **Widen the peer range** to admit the new seams: `^0.1.0-rc.6 || ^0.1.1-rc.1 || ^0.1.2-alpha.4 || ^0.1.3-alpha.2 || ^0.1.5-alpha.1` (old clauses stay for install compatibility; hosts on pre-0.1.3 tuples reject the new shape at compress time and must pin a pre-fix release).
4. **Move the `@deepseek-ai/*` devDep baseline to 0.1.5-alpha.2 (exact, full closure pinned)** — the regression tests must run against a session that *validates* the new shape; an old-line session would silently accept the old fields and prove nothing. Also required: `assistant/message` fixtures now carry `stream: []` (the 0.1.5 token-meter reads it), and the e2e harness dropped its double `SessionProjectionRegistry` registration (the agent-loop-testkit already mounts it) and switched to `session.snapshotEvents()` (the `events` getter is removed).
5. **Regression test** (`tests/region.test.ts`, "M5: the checkpoint replace op speaks the current SurfaceOp protocol"): runs a real compaction transaction and asserts the checkpoint event's `surfaceOp` is exactly `{ op:'replace', startSeq:1, endSeq:4 }` with three keys and no `start`/`end` — the pinned session IS the validator (the old shape throws at append).

## Verification

- `npm run typecheck` — clean (strict TS).
- `npm test` — **215 pass / 0 fail** (214 baseline + the new regression test; 4 pre-existing tests updated for the 0.1.5 line: `stream`-aware meter fixtures, `snapshotEvents` accessor, the #124 host-helper test now pins the *fixed* 0.1.5 helpers which read `session.eventAt`).
- `npm run build` — clean.
- `npm run test:e2e` — **PASS**: the real in-process host (dsh-agent-loop 0.1.5-alpha.2) drives compress end-to-end; "durable replace node landed", summary/projection/decompress assertions all green.


ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework-aio)

[bot] 🏷 Review started. Checking out the PR branch in a local worktree now; I'll run the full gate (`typecheck` / `test` / `build` / `test:e2e`) to independently verify the claims, then review the diff for cleanliness and protocol correctness before giving a verdict.

hairyf:
我看到了另外的修复 PR，如果需要该 PR 可随时重新打开，感谢

Issue #141, opened 2026-09-11
(bug) compaction/summary persists non-format members (tier/topic/kernelBlockId/…) — pre-2.0.9 v0 session logs become unopenable after host upgrade


EkkoXy:
## Summary

On DSH **2.0.9** (harness core `0.1.5-rc.1`) the host's **frozen released-v0 session reader** (`@deepseek-ai/dsh-session-format-v0-to-v1`) refuses to migrate any v0 session log whose events carry members outside its exact allow-list. `billion-context-dsh` persists several plugin-private members into `compaction/summary` data, so **every session that ran a compaction before the host upgrade can no longer be opened**:

```
failed to observe session session-...: @deepseek-ai/dsh-session-format-v0-to-v1 refuses this format v0 Session:
compaction/summary 38224 data has unexpected member tier;
source v0 artifact remains unchanged (raw log: .../session.jsonl.zstd)
```

This is a **different code path from #136 / PR #137** (that one is the runtime `surfaceOp` dialect). #137 does not touch these members — they are still written on `main` today: `src/region.ts` contains `tier:` ×5, `topic:` ×2, `kernelBlockId` ×29, `directMessageIds` ×11, `effectiveMessageIds` ×11, `parentBlockIds` ×15; `src/state.ts` adds more.

## Environment

| | |
|---|---|
| billion-context-dsh | 0.2.21 (still present on `main` post-#137) |
| host | DSH Desktop 2.0.9, harness core `0.1.5-rc.1` |
| rejecting reader | `@deepseek-ai/dsh-session-format-v0-to-v1@0.1.5-rc.1` |

## Mechanism

`lib_index.js:60` defines the released-v0 payload disposition for `compaction/summary` — admitted members are exactly:

- required: `compactionId, summary, shadowedRange, shadowedSeqs, shadowedTokenCount, provider, model`
- optional: `sourceCommandId, maxTokens, usage, rawOutput, llmStreamCall`

and `lib_index.js:260`:

```js
if (unexpected !== void 0) throw new SessionFormatError(`${label} has unexpected member ${JSON.stringify(unexpected)}`);
```

The engine appends `tier`, `topic`, `kernelBlockId`, `directMessageIds`, `effectiveMessageIds`, `parentBlockIds` in the same data object. The host rejects the whole file on the first offending member, so only `tier` is reported.

## Impact

[...]

PR #143, opened 2026-09-11
(feat) frame model-written summaries, record verified readings, slim nudges (drop unchanged-injection elision)

ranxianglei:
Supersedes #132 (original design by @TheTouYu). Lands the maintainer-reviewed subset of the injection-governance work: **B1** summary source framing, **B3** verified readings, **B6** slim nudge. **B2** unchanged-injection elision is deliberately dropped (evidence below).

## Problem

Three separate issues in how compression output is written and read:

1. Model-written compression summaries carry obligation-like sentences distilled from compressed user messages. Nothing marks them as model-written, so a model reading a summary back can execute stale secondhand directives as if they were current user words.
2. `compress` accepted `verifiedReadings` (files/sections actually read and checked) but persisted them as top-level members of the `compaction/summary` event — the frozen v0 log reader validates an EXACT member allow-list and throws on unknown members, so those writes brick pre-upgrade session logs (same class as issue #141). And nothing ever read them back: a durable datum nobody reads is pure liability.
3. Every nudge body repeated four guidance sections (compression philosophy, compress rules, tier-2/tier-3 rules) that already live once in the system prompt — pure waste on every single nudge.

## Why B2 (unchanged-injection elision) is dropped

Audited against the real host packages (`@deepseek-ai/*` 0.1.5 line):

- **No host channel re-injects a frame while an identical copy stays visible.** `RuntimeContextProjection` in dsh-agent-loop skips when the retained snapshot text is identical (`if (this.retained?.text === snapshot) return;`); agent-instructions re-inject only when the current copy disappears from the surface (rule 16 live audit). So the elision saved zero bytes by construction — the author's own live audit recorded 0 bytes saved and 2 content losses.
- **The stub would steal #93's newest-row pin.** The spread-stub kept `source.changes[].scope` (the frozen v0 reader validates exactly `{kind:'agent-instructions', form:'instructions', changes[]}`), so the stub becomes the guarded newest row while the FULL copy degrades to a compressible stale row — content loss with no re-injection.
- **The whitelist matched at most one real frame kind**: `skill-catalog`/`runtime` kinds exist nowhere in the published seam; real runtime-context frames carry `kind:'plugin'` + `plugin:'@deepseek-ai/dsh-system-prompt'`.

Full evidence and the conditions under which B2 would be redesigned: `docs/injection-governance-design.md` §2.

## Changes

Rebased onto current main (`baa8834`); the original commit is preserved verbatim in history, with one review-fix commit on top.

- **B1** — every summary is prefixed ONCE at creation with `[Model-written summary — not user words; re-verify any obligations before relying on them]` (English: tool output speaks the English vocabulary), written identically into the `compaction/summary` event AND its checkpoint node (one `framedSummary`; divergence would split the decompress header from the projected text). Projection-time framing stays as an idempotent safety net for legacy blocks only. The private `isCompactionCheckpoint` predicate is deleted in favor of the existing `isCheckpointNode`. Real user frames are never framed.
- **B3** — `verifiedReadings` now ride the rawOutput block ledger (seventh field of `encodeAcpBlockLedger`, omitted when empty) — never top-level. Read face `verifiedReadingsOf(event)` decodes embedded ?? legacy top-level, `Array.isArray`-guarded, never throws; `rebuildBlockLedger` surfaces them on `AcpBlockLedgerEntry`; the compress result echoes `, verified: <a; b>` per landed block (non-empty only).
- **B6** — `stripNudgeGuidance` removes the four guidance sections from nudge text on BOTH the kernel default path and the template path (host-overridden slots included). Measured nudge body ≈224 B ≤ 300 budget. Tests pin substrings of REAL rendered output (trigger frame present, `Context breakdown:` present, guidance absent) instead of constant round-trips.
- **Tests** rewritten against the real dsh-session seam API (`sessionEventsOf`/`eventAtOf` — the 0.1.5 line has no `session.events` getter, which made the original test file crash at runtime); the M3 decompress test now expects the framed summary header.
- **Docs synced in the same change**: new `docs/injection-governance-design.md`, README.md / README.en.md feature notes, AGENTS.md module map + new hard-won rule 17, docs/README.md index.
- `dist/` intentionally untouched (dist-bot refreshes after merge).

## Verification

- `npm run typecheck` — clean
- `npm test` — 273/273 pass
- `npm run build` — clean
- `npm run test:e2e` — all scenarios pass

@TheTouYu — thanks for the design; your original commit is preserved as-is in the history of this branch. If a future host genuinely re-injects over visible copies, B2 can come back redesigned against the real shapes (stub must not carry `changes[]` identity and must classify as engine metadata) — the decision record in the design doc spells out the conditions.


Tyan66666:
**Maintainer verification — three docs/process findings fixed in `7b2de21`**

I merged this branch onto the current `main` (`f7ec0e2`, i.e. after #130) in a worktree — no conflicts — and ran the full gate on the merged tree: `tsc --noEmit` clean, **295/295 unit tests pass**, `npm run build` clean, `npm run test:e2e` **PASS** (including the wire-level checks added in #126). The three feature points check out:

- **B1** frames the same text in both the event (`src/region.ts:481`) and the checkpoint node (`:494`); only checkpoint nodes get framed at projection time (`src/messages.ts:141`), so real user turns are never framed.
- **B3** keeps `verifiedReadings` inside the `rawOutput` payload (`src/block-ledger.ts:51/76/115`), so the top-level members of `compaction/summary` are unchanged — no repeat of the #141 v0-reader rejection.
- **B6** strips the guidance blocks on both the kernel path (`src/nudge.ts:319`) and the template path (`:467`).

I also re-checked your reason for dropping B2 and it holds: the host's `RuntimeContextProjection` returns early when the snapshot text is unchanged (`dsh-agent-loop/lib/index.js:299-337`), so eliding an unchanged injection saves 0 bytes by construction.

Three problems turned up, all documentation or process — no runtime defect. I pushed a fix for them instead of leaving them to you:

1. **The docblock on the function that does the stripping described the old behavior.** `adaptKernelNudgeToSeq`'s docblock said "Everything else (frame, philosophy, breakdown, HOW_TO_COMPRESS_RULES, tier rules, tip) stays kernel verbatim", but the function's first statement calls `stripNudgeGuidance`. Rewrote it — and the inline tier comment — to describe the real order: strip first, then adapt the refs; only the frame, the context breakdown, the tier line and the tip stay kernel-verbatim.

2. **B6 is a host-side workaround, but it carried no label and no tracking row.** Stripping sections out of already-rendered nudge text is a local post-process of kernel output; the clean fix belongs upstream as an acp-kernel option or parameter that renders the nudge without the guidance blocks. AGENTS.md rule 11 asks for an `UPSTREAM:` comment plus a tracking row with a removal gate, so I added the comment above `GUIDANCE_BLOCKS` in `src/nudge.ts` and a row in `docs/dsh-porting-verification.md` whose removal gate reads: delete `stripNudgeGuidance` once the kernel offers that option. That row currently notes that **no upstream issue has been filed yet** — that is the next step.

3. **`docs/configurable-prompts-design.md` still documented the pre-B6 templates.** Five places — the R1 entry, the K2 table row, the P1/P2/P3 rows and the `NudgePrompts` code excerpt — claimed the normal/emergency templates embed `{philosophy}` and that `guidance` defaults to the kernel `HOW_TO_COMPRESS_RULES` verbatim. Added a v8 revision note that states the correction explicitly, updated those rows, and mirrored the same three interface comments in `src/prompts.ts`.

Verification after the fix: `tsc --noEmit` clean, 295/295 tests, build clean, e2e PASS. The change is comments and documentation only, so no behavior moved. Pushed as a fast-forward to `2026-09-11_injection-governance-b1-b3-b6` (no force).

PR #144, opened 2026-09-11
(feat) image and file blocks stay visible to the engine — placeholder projection + host-meter media price (#117)

### LINKED ISSUE #117: [multimodal] image block 在 ACP 投影层完全不可见:同一份 acp_status 两套标尺、range 表对图片区间记 0 token、图片-only 事件无 bare ref 且不受内核保护
来源: ranxianglei/billion-context#582 第二轮扫描(用户指定多模态专项)。grep 确认 `src/` 全目录零处 image/multimodal/mediaType/base64 引用,而宿主侧已有多模态管道:`@deepseek-ai/dsh-llm` 定义 `ImageBlock { type: 'image' }`(types.d.ts:49-55,注释"only user content carries images today")并有共享的递归 image walk 供 capability gating / text-only serialization / compaction 三类策略使用——平台能带图,引擎视而不见。

## 具体断点(均已在代码中验证)
1. **投影丢弃**:`extractText`(src/messages.ts:27-41)只取 `type:'text'` 块与嵌套数组,image 块既非 text 也无 content 数组 → 静默丢弃;`projectEvent`(messages.ts:119-175)因此对 image-only user message 产出空数组。后果链:
   - **无 ref**:image-only 事件在 kernel 视图里不存在 → 永不获得 mN/bare-seq ref;
   - **无边界资格**:`hasPlainRef`(src/region.ts:110-132)对 user/message 要求 `extractEventText().trim().length > 0`,image-only → false → 模型把图片消息当 compress 边缘时 `resolveSurfaceRange` 会向内收缩或向外扩张吞邻接内容,或直接抛 "no tool-pairing-balanced range"——报错信息与真实原因(图片)无关,模型只能盲试;
   - **内核保护盲区**:`handleCompress` 依赖 kernel applyCompression 的 recent/last-user 保护(src/tools.ts:416-419 注释),而保护基于传入的 coreMessages——image-only 的最后一条 user message 不在其中,**可被压缩掉**;dsh 自己的 last-user 保护只在 range 表路径(src/region.ts:850-856),不覆盖工具直接请求路径。
2. **双标尺混印**(bc#386 "双估计器混印不标尺" 的同族,但 dsh 里是结构性的):nudge/emergency 闸门用 `resolveTokenCount` → `projectedTokens`(provider 口径,**含**图片 token),而 acp_status breakdown(`buildStatusReport` + `defaultCountTokens`,src/tools.ts:742)、nudge range 表(`buildCompressibleSeqRanges` 逐事件 `defaultCountTokens(extractEventText)`,src/region.ts:877)、`/acp status` 的 blocks 统计全是 text-only 口径。一份报告里 "Nudge: ACTIVE — usage 92%" 和 breakdown 百分比来自两个世界;图片密集会话里偏差不是 CJK 级的百分之几,而是整张图的 token 数(单图可达千级 token 记为 0)。
3. **压缩优先级误导**:range 表按 tokens 排序引导模型先压最大区间,图片密集区间显示 ~0 token → 模型优先压掉的恰恰是最不占空间的区间,真正的空间大户(图)永远排不上号。
4. **召回静默丢失**:`decompress`/`search_context` 走 `extractEventText`,恢复出的 block 文本里图片上下文整体消失,search 索引也搜不到图片相关描述(若图在文字里有引用则部分幸存)。

## 对照
- bc 侧同类问题的处理:#488(图片字节缺失于 preflight gate)、#496(image-estimate false 502,forward-once-then-learn)、#386(统一估计器标尺);dsh 的 #103/#107 已修了 shadow-price 的图片计价通道(宿主 meter 口径),但**引擎自身的投影/标尺/边界/保护四层都没接图片**。

## 建议(分层,供 owner 拍板)
- 最小修复(不改投影):① acp_status 显式标注两套标尺(provider-anchored vs text-only estimate);② `hasPlainRef`/`buildCompressibleSeqTokens` 对含 image 块的事件用宿主 meter 价(node.tokens/heuristicTokens,#107 同源)而非 0;③ `handleCompress` 路径补 dsh 侧 last-user 保护,覆盖 image-only 事件。
- 完整方案:`projectEvent` 为 image 块产出占位 CoreMessage(contentType 保留信息、text 记 `[image …]` 占位符+meter 权重),让 ref/保护/range 表/search 全链路可见;需与宿主 text-only serialization 策略对齐(纯文本网关路由下图本就不发,投影占位不影响出站)。
- 测试:现有 tests/ 无任何 image 用例,补 image-only user message 的投影/ref/边界/保护/标尺五连用例。

Tyan66666:
## Problem

A screenshot session carries information the engine cannot see. `image` and `file` content blocks have no characters, so `extractText` (`src/messages.ts`) dropped them and every text estimator priced them at 0. Four visible failures followed from that one omission (issue #117):

1. The message had **no kernel ref at all**, so it could not act as a compress boundary: `hasPlainRef` (`src/region.ts`) requires non-empty event text, so `resolveSurfaceRange` shrank inward past it and swallowed neighbouring content — or failed with `no tool-pairing-balanced range`, a message that names an unrelated cause.
2. The kernel's recent/last-user protection never saw it. On a screenshot session the last user message is often image-only (the question *is* the picture), and it could be compressed away.
3. The nudge range table ranked picture-heavy spans at ~0 tokens, so the model compressed the cheapest-looking content first.
4. `decompress` / `search_context` restored the text but silently lost the image context.

## Cause

Nothing upstream is buggy. `acp-kernel`'s token boundary is a `(text: string) => number` function, so the kernel cannot price media either; and DSH's own answer for files is to project them to handle text ("request assembly projects every occurrence to deterministic handle text (name, byte size, read-only saved path)" — dsh-llm `FileBlock` docblock). The engine had no equivalent step for `image` / `file` blocks.

## Fix

- **`src/messages.ts`** — `extractText` renders a deterministic placeholder for `image` / `file` blocks at any nesting depth (including inside `tool-result` content): `[image image/png shot.png 800x600 4.2KB]`, `[file notes.txt 1.0KB]`. One change restores the ref, the boundary and the kernel protection, and it mirrors the host's own file-handle projection instead of inventing a second convention. `attachmentPlaceholder` returns null for a malformed or foreign attachment — it contributes nothing and invents nothing.
- **`src/host-tokens.ts`** — `mediaPriceViaMeter(session, ctx)` reads the live token-meter's routed structural surcharge (`nodes[seq].imageStructuralTokens + fileStructuralTokens`) into a seq→price map; a missing or throwing meter yields an empty map.
- **`src/region.ts`** — `compressibleSegmentsOf` takes `mediaPriceOf` and prices only media-bearing seqs; ranges carry `images` / `files` counts and render `[+N images | +M files]`.
- **`src/nudge.ts` / `src/prompts.ts`** — `mediaPriceOf` is threaded through `rangeTable`, `buildNudgeText`, `adaptKernelNudgeToSeq` and `renderNudgeFromTemplates`; the default range-row template gains a `{media}` slot.
- **`src/tools.ts`** — `acp_status` adds one line explaining the two calibers when media is on the surface (the pressure line is provider-anchored; the breakdown above it is a text-only estimate), and leaves the kernel report untouched otherwise.

Deliberately lazy: the meter is measured at most once per nudge, and never for a session with no media on its surface — issue #110's per-step cost is untouched. The shadow-price writers (`shadowedTokenCount`) are not touched: display currency and event currency stay separate (AGENTS.md rule 12).

## Verification

- `npm run typecheck` clean; `npm test` **324/324** (318 existing + 6 new); `npm run build` and `npm run test:e2e` PASS.
- New `tests/media-visibility.test.ts` covers: placeholder shape, nesting depth and malformed attachments; nested `tool-result` counting; ref + boundary restoration; range pricing marked **and added** (the assertion compares the same span priced at 1500 vs 0 — an initial `>= 1500` still passed with the surcharge dropped, so it was replaced); meter degradation (absent / throwing); the `acp_status` caliber note.
- Mutation-verified in both directions: remove the placeholder branch → tests fail; drop the media price from the token sum → test fails; remove the caliber note → test fails.
- Fixture note: the range table protects the last 5 surface nodes plus the last real user turn, so a 4-node session yields zero compressible ranges — the media fixture uses 15 nodes with the screenshot in the old half.

## Docs

AGENTS.md gains hard-won rule 19 plus module-map rows; `README.md` / `README.en.md` gain a how-it-works row; new `docs/media-visibility-design.md`; `docs/README.md` index row; `docs/dsh-porting-verification.md` appendix row 11.

Also fixed here: AGENTS.md had **two rules numbered 17** (the settings-layer rule and the summary-framing rule). The summary-framing rule is now 18, this PR's rule is 19, and the module-map references follow. The settings rule keeps 17.

Closes #117.


Tyan66666:
**Review follow-up pushed — the media price was reading fields that do not exist**

An independent review pass over the first commit found a real bug in the pricing half; fixed in `487b2e0`.

- `mediaPriceViaMeter` read `nodes[seq].imageStructuralTokens` / `fileStructuralTokens`. The published `TokenSurfaceNode` is `{ seq, tokens, heuristicTokens }` — neither field exists — so the media price was always **0 in production** and the "picture-heavy spans no longer look free" claim was false. The tests hid it: the stub meter invented the same two fields, so the suite was green against a type the runtime never produces.
- It now reads the surcharge the route actually charged — the difference between the node's two **real** prices (`tokens - heuristicTokens`) — and adds the fixed structural price the host heuristic charges for a media reference (`hostMediaStructuralPrice`, a mirror of the host's `estimateStructuralBlock`). The addition is load-bearing: no adapter declares a visual price today, so the difference alone is 0, and an absent surcharge must not read as a free image.
- Test hardening: the meter test drives the real node shape and asserts the difference (a wrong field name now turns red instead of passing quietly). Two new cases cover an image-only **last user turn** — real turn, kernel ref, never offered for compression — and `search_context` still finding the screenshot after its span was compressed. Mutations verified: dropping the fallback, or the difference, each turns a test red.
- Docs corrected where they described the old field reads (README zh/en, `AGENTS.md` rule 19, `docs/media-visibility-design.md`, verification appendix), plus the remaining limitation: `acp_status` drilldown rows stay text-priced, because the kernel's `countTokens` callback accepts text only.

Gates on `487b2e0`: `npm run typecheck` clean, `npm test` 326/326, `npm run build`, `npm run test:e2e` PASS.

ranxianglei:

**Confirmed the bug, reviewed the fix, gates green — plus one small doc/code nit I prepped a fix for.**

**1. The phantom-field bug is real (verified independently).**
The public node `measure()` returns is `TokenSurfaceNode = { seq, tokens, heuristicTokens }` (`@deepseek-ai/dsh-token-meter/lib/types/types.d.ts:39`). There is **no** `imageStructuralTokens`/`fileStructuralTokens` on it — those fields live only on the internal `MeterSurfaceNode` (`surface-fold.d.ts`). So the first commit's `(node.imageStructuralTokens ?? 0) + (node.fileStructuralTokens ?? 0)` evaluated to `0` for every node in production, and the stub meter that invented those same fields kept the suite green — a textbook case of a test fixture that doesn't mirror the runtime shape. Your diagnosis is correct. (Honest note: my own verification checked the attachment-ref types feeding the placeholder half but not the token-meter node shape feeding the price half — that's the gap that let the phantom read through.)

**2. Your fix (`487b2e0`) is sound.**
Reading the routed surcharge as `tokens − (heuristicTokens ?? tokens)` is exactly the right signal: it's zero on every adapter that declares no visual price (all of them today), and `compressibleSegmentsOf` adding `hostMediaStructuralPrice(...)` on top means a picture-bearing span is never priced as free. On a priced route the two terms collapse cleanly (`visual − structural + structural = visual`). It stays display-only and never feeds a `shadowedTokenCount` claim (rule 12). The `messages.ts` refactor (shared `contentBlocksOfEvent`) keeps the count and block accessors walking the same path, and the hardened meter test driving the real `{ seq, tokens, heuristicTokens }` shape is the right guard — a wrong field name now turns red instead of passing quietly.

**3. Gates — clean run in a fresh environment (head `487b2e0`):**
- typecheck: ✅ exit 0
- test: ✅ **327/327** (your 326 + the one in §4)
- build: ✅ tsup success (`dist/index.js` 262.55 KB)
- e2e: ✅ all 4 scenarios PASS

On the errors in your pasted output — both look environmental, not present in the committed head:
- `region.ts:1345 … Expected ";" but found "else"` — that's not in `487b2e0`; line 1345 there is a comment and `tsc --noEmit` is clean. It reads like a mid-edit working-tree state when you captured that run (the same log also carries a clean `326/326`).
- `ERR_MODULE_NOT_FOUND: @deepseek-ai/dsh-session-format-v0-to-v1` — that **is** a declared devDep (`package.json:69`) and present in a clean install; `block-ledger.test.ts` passes here. A stale/partial local `node_modules`; a clean `npm ci` clears it.

**4. One remaining nit (doc/code mismatch) — fix prepped, but I can't push to your branch.**
AGENTS.md rule 19 documents the combined row as `[+N images | +M files]` (and the sibling `[tool X% | text Y%]` bracket uses a pipe), but `mediaSuffixOf` in `src/nudge.ts` still joins with a comma, so a mixed row renders `[+2 images, 1 file]`. No test covered the mixed case — the single-image fixture can't exercise the separator. I aligned the code to the documented pipe form (per-part `+`; single-type rows unchanged) and added a combined-case regression test asserting `/\[\+1 image \| \+1 file\]/`. The full gate set passes with it (that's the 327th test above).

I don't have write access to `Tyan66666/billion-context-dsh` from this account (push → `403 … denied to ranxianglei`), so I couldn't land it on `pr-117`. If you agree with aligning to the docs, the core hunk is:

```diff
-  if (range.images > 0) parts.push(`${range.images} image${range.images === 1 ? '' : 's'}`)
-  if (range.files > 0) parts.push(`${range.files} file${range.files === 1 ? '' : 's'}`)
-  return ` [+${parts.join(', ')}]`
+  if (range.images > 0) parts.push(`+${range.images} image${range.images === 1 ? '' : 's'}`)
+  if (range.files > 0) parts.push(`+${range.files} file${range.files === 1 ? '' : 's'}`)
+  return ` [${parts.join(' | ')}]`
```

(plus a new test in `tests/media-visibility.test.ts`). Alternatively, if you'd rather keep the comma, flip the wording in rule 19 instead — your call.

**Bottom line:** the pricing fix is correct and the head is green across all four gates. Everything's done except your decision on the separator direction in §4.

*中文小结:* 独立核实了「媒体价格读了不存在的字段」是真 bug(公开的 `TokenSurfaceNode` 只有 `{seq,tokens,heuristicTokens}`,那两个字段只存在于内部 `MeterSurfaceNode`),你在 `487b2e0` 的修法(读 `tokens−heuristicTokens` 的路由差值 + 叠加固定结构价 `hostMediaStructuralPrice`)正确,四道门禁在干净环境全部通过(typecheck 干净、测试 327/327、build、e2e 全过);你贴的两处报错是本地环境问题(`region.ts:1345` 语法错是编辑中间态、缺失的 devDep 干净安装即有);另发现 range 行分隔符代码用逗号而文档(rule 19)规定竖线,我已备好对齐修复并加了回归测试,但没有你这个仓库的写权限推不上去,给出 hunk 由你定方向——确认分隔符写法后即可合并。

Tyan66666:
Thanks — real review, and the nit was right. Landed your hunk in 71c64d4.

**What changed:** `mediaSuffixOf` (`src/nudge.ts`) now puts the `+` on each part and joins with ` | `, so a mixed span renders `[+1 image | +1 file]` as rule 19 and the design doc describe.

**What I added on top:** a regression test that builds a span carrying one image AND one file and asserts the *rendered row text* (`/\[\+1 image \| \+1 file\]/`), not just the row struct. Your §4 said "no test covered the combined case" — it was slightly worse than that: no test asserted the rendered suffix at all, only `range.images` / `range.files`, which are separator-independent. That is exactly why the drift survived review, and why the single-type row passed with either separator. Mutation check: putting the separator back to a comma turns the new test red (and the single-type assertion too, since the per-part `+` and the join are one change).

**Your two environment diagnoses, confirmed:**
- `src/region.ts:1345 … Expected ";" but found "else"` — an editor linter's autofix (`no-negation-else`) rewrote a valid two-line `if (…) … / else …` into an invalid one-liner, and it resolved worktree paths back into the main checkout. It hit me 4+ times today. I have removed that plugin, so it cannot recur.
- `ERR_MODULE_NOT_FOUND: @deepseek-ai/dsh-session-format-v0-to-v1` — correct, stale/partial `node_modules`; it is a declared devDep (`package.json:69`) and `npm ci` clears it.

**Gates after the change (worktree, this branch):** `typecheck` ✅ · `test` ✅ 327/327 · `build` ✅ (`dist/index.js` 262.55 KB) · `e2e` ✅ 4 scenarios.

Also worth recording for whoever reads this thread later: your review caught the phantom-field bug by reading the published types, which is precisely what my own tests failed to do (they invented the same two fields, so the dead path stayed green) — the test now drives the real node shape (`{ seq, tokens, heuristicTokens }`) and asserts the difference. Credit for the review and for the test you could not push (403, no write access on this fork) will go in the release notes.

ranxianglei:

Verified on the branch — `71c64d4` is exactly the agreed hunk plus a better test, and all four gates are green when I re-ran them on that commit.

**Landed diff (`487b2e0..71c64d4`):** the three changed lines in `mediaSuffixOf` are the agreed hunk verbatim (per-part `+`, join with `' | '`) with a clearer WHY comment than mine. The test generalizes `buildMediaSession` to a `'mixed'` fixture and adds `appendMixedToolResult` in the real DSH shape (`{ type:'tool-result', toolCallId, content:[IMAGE_BLOCK, FILE_BLOCK] }`).

**Your point about the test is the real takeaway.** Asserting `range.images`/`range.files` was separator-independent — it passes with either `,` or ` | `. Your new assertion runs against the *rendered* `rangeTable` text (`/\[\+1 image \| \+1 file\]/`) and adds a `doesNotMatch(/\+1 image,/)` guard, so the comma form can't quietly come back. That's precisely why the drift survived: nothing ever looked at the string the model actually reads. This is the more durable form of the fix — good catch.

**Gates re-run by me on `71c64d4`:** typecheck ✅ exit 0 · test ✅ **327/327** · build ✅ `dist/index.js` 262.55 KB · e2e ✅ PASS. Matches your numbers.

**On the two env diagnoses:** both check out. The `no-negation-else` autofix rewriting valid `if/else` into an invalid one-liner is a nasty one — glad that plugin is gone; that "the file changed under me" class of failure is exactly what makes a red gate look like your own mistake. The stale `node_modules` one is straightforward.

Nothing left on my side — the PR is ready to merge whenever you are (leaving the merge button to you). Thanks for landing it clean and for the credit; the *"read the published types, not the fixture you invented"* lesson is worth keeping for every seam-bound field.

*中文小结:* 已在分支上核实 `71c64d4`(即商定的三行 hunk 逐字落地 + 你更强的「断言渲染后行文本」回归测试),并亲自重跑四道门禁全部通过(typecheck 干净、测试 327/327、build 262.55 KB、e2e PASS);你指出的「旧测试只断言与分隔符无关的 struct 字段、从不断言模型真正读到的那行文字」正是这个漂移能溜过 review 的根因,新测试已把它堵死;PR 我这边无遗留,等你合并即可。

PR #153, opened 2026-09-17
(fix) register agent/request-error context-overflow recovery — emergency compaction + retry instead of rethrowing unretried

Premshay:
## Problem

The ACP engine registers one `agent/pre-step` listener (the nudge) and **zero** `agent/request-error` listeners. The host loop retries a failed model request only when a listener answers the `agent/request-error` waterfall with `{ kind: 'retry' }` — otherwise it rethrows the provider error. ACP compositions disable `compaction-basic` (the only engine that answered it), so a real provider `CONTEXT_WINDOW_EXCEEDED` fails the turn unretried.

## Cause

`AcpCompactionEngine` copied the pre-step half of the host's automatic triggers but never the overflow half. The porting doc already requires both listeners — `docs/dsh-porting-verification.md`, architecture fact 2: the engine should copy compaction-basic's `agent/pre-step` (pressure) **and** `agent/request-error` (context-overflow recovery, `CONTEXT_WINDOW_EXCEEDED_CODE`). The doc is aspirational; the shipped engine registers only the pre-step listener and returns `null` from `compactIfNeeded`/`compactNow`.

## Fix

Register `agent/request-error` in the engine (`src/index.ts`), mirroring the host's overflow branch while keeping the model-driven design: **pressure stays nudge-only; the overflow path is the only automatic action.**

On a provider-confirmed `CONTEXT_WINDOW_EXCEEDED`:

1. `compactForOverflow` picks the **largest eligible block**: one kernel range spanning the whole surface, split by the existing guards (`buildCompressibleSeqRanges` — protected tail, last real user turn, current instruction rows, checkpoints, the system node), ranked by tokens in the same vocabulary as the nudge range table.
2. It lands the normal durable transaction with a **fixed engine-written marker summary** — no LLM call (the provider just rejected the request), `framed: false` so the model-written-source prefix is not stamped on engine-written text. The originals stay in the append-only log: `search_context` still indexes them, `decompress` restores them, and the model can re-run `compress` over the marker later to write a real summary.
3. It returns `{ kind: 'retry' }`, so the loop retries from the replacement surface.

Safety rails mirror the host: the recovery budget is `maxOverflowRetries` (new engine config, default **1** — the host's own default; validated non-negative integer), and it resets when the agent makes progress (an assistant message lands) or returns to idle. When nothing eligible exists, the listener hands the original error back untouched — best-effort, never a phantom block.

Supporting changes: `runCompactionTransaction` gains `framed: false` (`src/region.ts`); `meterMediaPriceResolver` is exported (`src/nudge.ts`) so the emergency path prices media spans like the nudge table does.

## Verification

- `npm run typecheck` — clean.
- `npm test` — **340/340 pass**, including 10 new regression tests (`tests/overflow-recovery.test.ts`): ownership of the overflow (`{ kind: 'retry' }`), non-overflow passthrough, budget exhaustion, progress/idle reset, nothing-eligible passthrough, aborted turns, `maxOverflowRetries: 0`, and fail-fast config validation.
- `npm run test:e2e` — **PASS**, with a new `overflow-recovery` scenario: the fake LLM returns an HTTP 400 with DeepSeek's context-length wording, which `dsh-llm-deepseek` normalizes to `CONTEXT_WINDOW_EXCEEDED`; the real agent loop then shows `requests: text,text,text,error,text` — the failed request, the engine's emergency compaction (start/end paired, marker summary, `context-overflow recovery` topic), the loop's retry, and a completed conversation. The other four scenarios still pass unchanged.
- `dist/` is intentionally not committed (PR branches never carry it; the dist-bot refreshes `main` after merge).


ranxianglei:

[bot] 🏷 PR #153 review

**Diff cleanliness (first gate): clean.** 8 files, every one purpose-related: the new listener + `compactForOverflow` (`src/index.ts`), the `framed: false` opt-out (`src/region.ts`), the `meterMediaPriceResolver` export (`src/nudge.ts`), e2e error-turn support + the new scenario, and the regression suite. No `dist/`, no lockfile churn, no unrelated reformatting, and the head sits directly on the current main tip (`4ff9a85`, v0.2.23) — nothing to rebase.

**Design review — correct split, verified against the real seam contracts rather than the description:**
- Pressure stays nudge-only; overflow is the only automatic action — exactly what `docs/dsh-porting-verification.md` architecture fact 2 asks for and what the host's own compaction-basic does.
- `agent/request-error` payload `{ agent, turn, step, provider, failure, retryPolicy, signal }` and `RequestErrorAction = { kind: 'retry' } | undefined` (dsh-agent `runtime-types.d.ts:357` / `:101`) — the listener returns precisely these shapes; non-overflow failures pass through untouched.
- `CONTEXT_WINDOW_EXCEEDED_CODE` (dsh-llm `error.d.ts:18`) is what `dsh-llm-deepseek` (`index.js:1537`, via `isContextWindowExceededError`) normalizes the provider wording into — so the fake LLM's HTTP 400 body exercises the real adapter path in e2e, not a shortcut.
- Rule compliance: orphan stripping applies transitively because `buildCompressibleSeqRanges` calls `stripOrphanedSurfaceToolMessages` internally (`src/region.ts:1202`); shadow pricing reuses the rule-12 path verbatim (`shadowedSeqsOf` + `shadowedTokensViaMeter(session, shadowed, agent.ctx)`); the topic rides the rawOutput ledger (rule 15); `framed: false` correctly skips the model-written-source prefix for engine-authored marker text (rule 18).
- `preserveRecent: 5` matches the nudge table's default (`src/region.ts:1206`), so the emergency path ranks spans with the same geometry and token vocabulary the model sees in nudges.
- Safety rails (budget, progress/idle reset, nothing-eligible passthrough, abort handling, generation-guarded retry decision) are all present, and each has a dedicated test.

**Local reproduction — all claims verified:** `npm run typecheck` clean · `npm test` **340/340 pass** · `npm run build` OK · `npm run test:e2e` **PASS**, with the new scenario showing exactly `requests: text,text,text,error,text` (failed request → emergency compaction with paired start/end + marker summary + `context-overflow recovery` topic → loop retry → completed conversation) and the other four scenarios unchanged.

**Findings (small, author-side):**

1. **Explicit null/undefined `maxOverflowRetries` passes validation but silently disables recovery.** `resolveAcpConfig` validates `resolved.maxOverflowRetries ?? 1` (`src/index.ts:256`) but returns `resolved` unmodified, while the listener consumes `this.config.maxOverflowRetries ?? 0` (`src/index.ts:604`). An absent key is fine (the spread keeps the default 1 — that's what e2e exercises), but an explicit `null`/`undefined` value (an empty YAML entry, or `{ maxOverflowRetries: undefined }` from the JS API) sails through fail-fast validation and then yields budget **0** instead of the documented default **1**. Suggested one-liner right after the validation block:
   ```ts
   resolved.maxOverflowRetries = maxOverflowRetries
   ```
   (then line 604's `?? 0` becomes unreachable — switch it to `?? 1` too if you want belt-and-braces.)
2. **Docs gap — mandatory per dev standards §4 ("Feature work MUST document itself").** The new user-facing key `maxOverflowRetries` has no row in the README.md / README.en.md config tables and no mention in the AGENTS.md module map.
3. *(Hygiene, optional)* `overflowRetries` / `overflowSessions` are unbounded Maps keyed by Agent/Session; #121 bounded similar per-session caches with LruMap. Low risk (bounded by process lifetime) — your call.

CI had not produced check-runs on the head commit yet at review time (`mergeable_state: blocked`, required checks pending) — I'll track it here when it lands.

**Recommendation:** implementation is solid, contract-verified, and fully reproduced locally. Request findings 1 and 2 as small follow-up commits before merge; I can't push to the fork branch myself, so those need to land from your side. Once they're in (and CI is green), this is good to merge.

---
中文摘要：本 PR 补上了 provider 上下文溢出（`CONTEXT_WINDOW_EXCEEDED`）时的紧急压缩+重试恢复——此前 ACP 组合下该错误会直接失败、不重试，因为引擎只注册了 pre-step 监听而漏掉了 `agent/request-error`；代码设计与真实 seam 契约逐项核对无误，typecheck/340 单测/e2e（含新场景）本地全部复现通过。需作者补两处小改动：显式 null 配置值会被静默禁用恢复的边界修复（一行），以及 README/AGENTS.md 缺少 `maxOverflowRetries` 文档行；补完且 CI 转绿即可合并。

Tyan66666:
Thanks @Premshay — this is exactly the gap we had (disabling `compaction-basic` also removes the only responder to `agent/request-error`, so a confirmed overflow was rethrown with no retry). I rebased your branch onto current `main` (no conflicts) and pushed the follow-ups I wanted before merging, so the review history stays in this PR.

**Fixes found while reviewing**

1. `resolveAcpConfig` validated `maxOverflowRetries` but never wrote the normalized value back, so an explicit `maxOverflowRetries: undefined` in a composition row survived into `this.config` and the listener's `?? 0` read "never set" as "disabled" — recovery was silently off. Normalization now happens at parse time (`src/index.ts:257-265`) and the listener fallback is `?? 1` (the host default; `0` can only come from an explicit `0`).
2. `overflowSessions` was written but never deleted, so the map held every Agent/Session for the life of the process. Both terminal paths now release the bookkeeping: `agent/status` → idle, and an `assistant/message` landing (progress).
3. **The engine-written marker was being labeled a model-written summary** (the real defect). `projectEvent` frames every checkpoint node with `SUMMARY_FRAME_PREFIX` ("Model-written summary — not user words…"), so the overflow marker — engine text, not model text — reached the model carrying a false claim about its provenance, while the durable text/`decompress` header (no prefix) no longer matched it: the stored-vs-projected split rule 18 already has a note about. A caller-side `framed: false` flag cannot fix that (the read side cannot see a flag), so the exemption now lives inside the one choke point both the creation write (`prefixSummaryBlocks`) and the projection net go through: `withSummaryFramePrefix` recognizes engine text by content (`ENGINE_SUMMARY_LEAD` / `isEngineWrittenSummary` / `overflowMarkerSummary`, `src/messages.ts`). The two byte streams are then identical by construction and the `framed` option plus its call site are gone.
4. `preserveRecent: 5` is now pinned explicitly with a comment: it is `region.ts`'s default, and the automatic path must not silently drift when the advisory range table gets retuned later.

**Docs** (this repo requires behavior changes to land with docs): new `docs/overflow-recovery-design.md` (why automatic is acceptable only here, the selection mechanics, budget/reset, the framing rationale, known limits), `maxOverflowRetries` rows in `README.md` / `README.en.md` plus the "advisory, never imperative" bullet naming this single exception, `AGENTS.md` (design decision 3, module map, rule 18a), `docs/dsh-porting-verification.md` (architecture fact 2 / decision D4), `docs/e2e-harness-design.md` (new scenario row + verification log).

**Verification**: typecheck clean; `npm test` 381/381 (12 in `tests/overflow-recovery.test.ts`, including "projection text == durable text" and "the marker carries no `SUMMARY_FRAME_PREFIX`" — mutation-checked: removing the content exemption turns it red); `npm run build` OK; `npm run test:e2e` 5 scenarios 65/65 PASS. The new `overflow-recovery` scenario drives the real host loop end to end — `requests: text,text,text,error,text`: the scripted HTTP 400 "maximum context length" reply is normalized to `CONTEXT_WINDOW_EXCEEDED`, one emergency compaction lands (`starts=1 ends=1`, marker summary, durable replace node), and the retried request succeeds.

Your original commit stays in the branch, so authorship is preserved.

ranxianglei:

[bot] 🏷 Re-review of the rebased head (`c571415`) — verified, ready to merge

**All three original findings are correctly resolved:**
1. **Budget normalization** — `resolveAcpConfig` now writes the validated value back (`return { ...resolved, maxOverflowRetries }`, `src/index.ts:257-265`) and the listener fallback is `?? 1` with a comment pinning why it must stay 1. New regression test ("an explicitly undefined maxOverflowRetries keeps the host-parity default of 1") nails exactly the edge case from round 1.
2. **Bookkeeping leak** — both terminal paths now release the maps: `assistant/message` progress deletes the agent's budget *and* the reverse lookup (`src/index.ts` session/event handler), `agent/status` idle does the same. Covered by "overflow bookkeeping maps release their entries on both terminal paths".
3. **Docs** — complete: `maxOverflowRetries` rows in both READMEs, the advisory-bullet exception note, AGENTS.md (design decision 3, module map, new rule 18a), porting doc fact 2 updated to post-#153 state, e2e harness doc, and a proper `docs/overflow-recovery-design.md` (TL;DR → why automatic is acceptable only here → mechanics → framing rationale → known limits). This satisfies §4.

**The framing defect you caught is the real one, and the fix is in the right place.** My round-1 review verified the creation-side write but missed that the projection-time idempotent net would re-stamp the engine marker with the model-written prefix — exactly the stored-vs-projected split rule 18 warns about. Your resolution is the correct architecture: exemption by content in the single choke point (`withSummaryFramePrefix`, `src/messages.ts:205`), one shared literal (`ENGINE_SUMMARY_LEAD`) between writer and both framing sites so identity holds by construction, and the dead `framed` option removed rather than left as a second condition. The tests are genuine behavior pins, not tautologies: durable text free of `SUMMARY_FRAME_PREFIX` (line 87), projected text free of it (line 99), and projected == durable byte-equality (lines 102–105) — removing the content exemption turns them red, as claimed. One accepted trade-off worth noting for the record: a model-written summary that happened to *begin* with the exact marker lead-in would also be exempt — low risk (advisory frame lost, not data corruption), and no better option exists since the read side cannot see caller flags.

**Regression review vs the new base** (main moved ~14 commits during review: v0.2.24/0.2.25, DSH 0.1.7 seams, `/acp-prune` rename): the rebase landed cleanly on the current tip, the diff is still 16 purpose-related files (+667/−21, no `dist/`, no lockfile churn), and every pre-existing behavior the diff touches — creation-side framing (rule 18), orphan stripping (rule 8, transitively via `buildCompressibleSeqRanges`), shadow pricing (rule 12, `shadowedTokensViaMeter` unchanged), ledger topic (rule 15) — was re-checked and holds. `preserveRecent: 5` is now explicitly pinned with a drift-prevention comment, which is the right call for an unattended path.

**Local reproduction on `c571415`:** typecheck clean · `npm test` **381/381 pass** · build OK · `test:e2e` all 5 scenarios PASS, overflow-recovery showing `errorIdx=3, requests=5` (failed request → emergency compaction, start/end paired, marker + `context-overflow recovery` topic → retry succeeds), other four scenarios unchanged.

**CI:** all required checks green on `c571415` — `ci`, `pr-title`, `e2e` all `success`; `mergeable_state: clean`.

**One nit left (non-blocking):** `src/region.ts:498-500` still says *"The engine-written emergency overflow marker opts out (`framed: false`)"* — that mechanism was removed, and the sentence directly contradicts the next paragraph (and rule 18a) which describe the actual content-based exemption. Deleting those two lines fixes it; comment-only, zero behavior impact. Happy for it to ride along as a tiny commit before merge or land as a post-merge fixup.

Verdict: implementation, fixes, docs, and tests all check out; CI green. This is ready for a human merge whenever you are.

---
中文摘要：复审通过——上一轮提出的三个问题（预算归一化边界、簿记泄漏、文档缺失）均已正确修复并带回归测试；作者额外发现并修复了我初审漏掉的真正缺陷（投影侧幂等帧网会给引擎写的超窗标记错误盖上"模型撰写"前缀），修复落在唯一的收口函数里、以共享字面量按内容豁免，持久文本与投影文本字节级一致且有变异可红测试钉住。新基线上本地复现全部通过（typecheck 干净、381/381、e2e 五场景全过），CI 三项检查全绿、mergeable_state 为 clean，仅剩一处过时注释（region.ts:498-500 仍引用已删除的 framed: false）不阻塞合并，可以合并。

PR #156, opened 2026-09-19
(fix) let multi-tool-call assistant messages and empty tool results anchor compress range edges

lxckck:
## 问题(Problem)

长会话中,一段**已被完全消费**的历史内容(并行多工具调用轮次 + 一个空 tool result)永远无法压缩:任何触及该簇的 compress 调用——整段、分段、甚至单条消息的最小粒度——都报:

```
compress: billion-context-dsh: no tool-pairing-balanced range around seq 75..75
```

单条也失败说明该簇内不存在任何合法边界。实测约 8–10K 已消费内容永久滞留表面,且随相似轮次累积继续增长。

## 原因(Cause)

三层叠加(均指 src/,v0.2.23):

1. `region.ts` 的 `hasPlainRef` 边界谓词把"无裸 ref"与"不可作边界"划了等号,拒绝了两类真实形状:
   - **多工具调用 assistant 消息**:`projectEvent` 投影为 `seq#callId` 子 CoreMessage,没有裸 `seq` id;
   - **空 tool result**:`projectEvent` 返回 `[]`,不生成任何消息。
   `resolveSurfaceRange` 的收缩/扩张两轮因此都找不到合法切点,全部坍塌后抛 "no tool-pairing-balanced range"。并行工具调用是现代 agent 会话的常态,空结果(空 grep/glob 等)也不罕见,相邻即成死区。
2. `tools.ts` 的 `byRaw[String(seq)]` 精确查找无回退——即使边界配对平衡,ref 也无处可取。
3. 配对平衡本身(dsh-compaction 的 `toolPairingBalancedBefore/After`)没有问题。

## 修复(Fix)

1. **`src/region.ts`**:`hasPlainRef` → `anchorsRangeEdge`——多工具调用消息(有 call 或非空文本)与空 tool result 均可作边界;docblock 说明理由(空结果无内容可失,multi-call 经子 id 锚定)。
2. **`src/tools.ts`**:新增 `nearestRefForSeq` 三级回退,替换两处精确查找:① 精确 `String(seq)`;② `seq#` 前缀子 id(任一子消息锚定同一表面节点);③ 沿**表面**向前找最近带 ref 的活节点(空结果回退到前一 ref;沿表面走而非裸 id 空间,避免锚到已压缩消息)。
[...]

PR #157, opened 2026-09-19
(fix) let consumed parallel tool-call rounds compress — multi-call messages and empty results anchor range edges

### LINKED ISSUE #155: 已消费的并行工具调用簇永久不可压缩——多工具调用消息与空结果无法作为压缩边界
## 环境

- billion-context-dsh v0.2.23(dist 内联 acp-kernel 0.0.63)
- DeepSeek Harness 0.1.5 线(dsh-session 0.1.5)
- 场景:单会话长跑,大量并行多工具调用轮次(multi-tool-call assistant 消息)+ 少量空 tool result

## 现象

一段**已被完全消费**的历史内容(多轮 multi-tool-call assistant 消息 + 一个空 tool result)永远无法压缩:任何触及该簇的 compress 调用——整段、分段、甚至单条消息的最小粒度——都被拒绝:

```
compress: billion-context-dsh: no tool-pairing-balanced range around seq 75..75
  — narrow the range or consult acp_status for the current surface
```

单条也失败,说明不是范围切法问题,而是该簇内**不存在任何合法边界**。实测长会话中约 8–10K 已消费内容因此永久滞留表面,且随相似轮次累积会继续增长。

## 根因(源码定位,均指 src/,v0.2.23)

三层叠加:

1. **`region.ts` 的 `hasPlainRef` 边界谓词过度保守**(`cleanBefore`/`cleanAfter` 同时要求配对平衡 + 裸 seq ref)。两类真实形状被拒之门外:
   - **多工具调用 assistant 消息**:`projectEvent` 将其投影为 `seq#callId` 子 CoreMessage,不存在裸 `seq` id → `hasPlainRef` 恒 false;
   - **空 tool result**(content 文本为空):`projectEvent` 直接返回 `[]`,连消息都不生成 → 无 ref 可言。
   于是收缩第一轮找不到合法切点、坍缩后扩张第二轮也找不到,两轮全塌 → 抛 "no tool-pairing-balanced range"。而并行工具调用轮次在现代 agent 会话中是**常态**,空结果(如返回空的 grep/glob)也不罕见——两者相邻即成死区。
2. **`tools.ts` 的 ref 查找无回退**:`byRaw[String(seq)]` 精确查找失败即 undefined → 硬错误;即使边界配对平衡,ref 也无处可取。
3. 配对平衡本身(dsh-compaction 的 `toolPairingBalancedBefore/After`)**没有问题**——问题仅在谓词把"无裸 ref"与"不可作边界"划了等号。

## 修复思路(补丁已实现并验证,可发 PR)

1. `region.ts`:`hasPlainRef` → `anchorsRangeEdge`——多工具调用消息(有 call 或非空文本)与空 tool result 均可作边界;docblock 说明理由(空结果无内容可失,multi-call 经子 id 锚定)。
2. `tools.ts`:新增 `nearestRefForSeq` 三级回退,替换两处 `byRaw[String(seq)]` 精确查找:① 精确 `String(seq)`;② `seq#` 前缀子 id(任一子消息锚定同一表面节点);③ 沿**表面**向前找最近带 ref 的活节点(空结果回退到前一 ref;沿表面而非裸 id 空间走,避免锚到已被压缩的消息)。

**附带收益**:issue #38 的非单调 checkpoint 形状([8,2..6,7])中,中段起点(3..6)的扩张现在**止于 multi-call 节点**(它是合法边界),不再越过 checkpoint 触发 8..6 值反转——该回归从"必须拒绝"变成"直接命中正确范围"。

## 验证(补丁分支)

- 全量套件 332 通过(含 kernel-upstream),`npm run typecheck`、`npm run build` 干净;
- 实测:先前不可压缩的三个死区段(共 ~10.5K)在打补丁后的构建下一次压缩成功。

如果方向认可,我把分支推上来开 PR。

ranxianglei:
Closes #155

## Problem

In long sessions with many parallel multi-tool-call rounds, fully-consumed history becomes permanently incompressible: every `compress` call touching such a cluster — whole span, sub-segment, or even a single message — fails with `no tool-pairing-balanced range around seq X..Y`. Measured ~8–10K of consumed tokens stranded on the surface, growing as similar rounds accumulate.

## Cause

A range edge counted as valid only if it carried a bare `${seq}` kernel ref (`hasPlainRef`, src/region.ts). Two real shapes never do:

- **Multi-tool-call assistant messages** project to `${seq}#${callId}` sub-messages — no bare ref;
- **Empty tool results** (a grep/glob that matched nothing) project to no CoreMessage at all.

So shrink found no legal cut and expand found none either; and even where pairing balance held, src/tools.ts had no way to fetch a kernel ref for such an edge (`byRaw[String(seq)]` exact lookup only). Pairing balance itself was never the problem.

## Fix

- **src/region.ts**: `hasPlainRef` → `anchorsRangeEdge` — an edge must be tool-pairing balanced AND anchorable: user messages with non-empty text, assistant messages with calls or text (a multi-call node anchors through its sub-ids; safe as a START edge only, since its open calls make every cut after it unbalanced), and any tool result including empty ones.
- **src/tools.ts**: new `edgeRefForSeq` maps resolved edges to kernel refs with a three-tier fallback: bare `${seq}` → the node's own sub-id (first for a start edge, last for an end edge, in content order) → nearest ref-bearing live surface node toward the opposite edge, clamped to the span (so the kernel consumes exactly what the durable transaction shadows). A span with no messages at all keeps the existing "no assigned ref" error.
- **src/region.ts** `seqOfKernelRef`: strip the `#callId` suffix before parsing — kernel ranges bounded by multi-call sub-ids were silently dropped from the model's compressible range table (a third affected spot, not visible from the reported symptom).

Side benefit: on the non-monotonic checkpoint shape from issue #38 (`[8, 2..6, 7]`), expanding a mid-round request now stops at the multi-call node instead of crossing the checkpoint into a value-reversed span — that regression resolves directly instead of being rejected.

## Verification

- Repro probe over real dsh-session objects covering the three dead-zone shapes (cluster before an in-flight tail, single-call in-flight tail, cluster behind a spliced checkpoint): all threw the reported error before the fix, all resolve after.
- Full suite 334/334 green (4 new regression tests + 3 stale pins rewritten to the corrected behavior); `npm run typecheck` and `npm run build` clean.

中文摘要：修复了并行多工具调用轮次 + 空结果构成的永久不可压缩死区（边界谓词把「无裸 ref」误判为「不可作边界」），让多调用消息经子 id、空结果经配对平衡直接充当压缩边界，并补上 nudge 范围表丢失子 id 边界行的问题；已加回归测试，全量套件通过，可合并。


ranxianglei:

[bot] 🏷 **Review of PR #157 — verified independently, mergeable**

**What I checked**
- Branch `ranxianglei:2026-09-19_range-edge-anchors` @ `b36a1d7` (single commit, Closes #155); CI all green (`ci`, `pr-title`, `e2e` — all completed successfully on this exact commit).
- Independent local re-run in a fresh worktree of the branch: `npm run typecheck` clean, full test suite **334/334 pass**, `npm run build` clean. Matches the PR description's claims exactly.

**Code review**

The root-cause diagnosis is correct: `hasPlainRef` answered "does this edge project to a bare `${seq}` ref?" where it should have answered "is this a legal cut?" — conflating an id-dialect detail with the pairing question, which permanently dead-zoned two real shapes (multi-tool-call assistant messages projecting to `${seq}#${callId}` sub-messages; empty tool results projecting to nothing). Pairing balance itself was never the problem, as the report says.

The fix reads correctly at all three spots:

1. **`anchorsRangeEdge` (src/region.ts)** — an edge must be pairing-balanced AND anchorable. Multi-call nodes anchor through their sub-ids; the comment's claim that they can only ever be START edges holds (their open calls make every cut after them unbalanced — I verified the argument). Empty results anchor unconditionally: they project to no CoreMessage and carry zero characters, so anchoring loses nothing, and the host pairing helpers count them like any other result. The system-node exclusion stays explicit in `cleanBefore`/`cleanAfter` rather than incidental.
2. **`edgeRefForSeq` (src/tools.ts)** — three-tier fallback (bare `${seq}` → the node's own sub-id, first for start / last for end edges in content order → nearest ref-bearing live node toward the opposite edge, clamped to the span). The clamp guarantees the kernel consumes exactly what the durable transaction shadows — no over-consumption past either edge. A span with no messages at all keeps the existing "no assigned ref" error, now covered by a dedicated unit test (two adjacent empty results).
3. **`seqOfKernelRef` strips the `#callId` suffix** — the third affected spot the bug report missed: without it, kernel ranges bounded by multi-call sub-ids were silently dropped from the model's compressible range table (NaN → null), invisible from the reported symptom.

Extra checks I made beyond the diff:
- The new assistant text check uses `extractText(content)` where the old code used `extractEventText(event)` — verified these are byte-identical for assistant events (src/messages.ts:282), so there is no hidden behavior change in that line.
- Docs synced in-PR as required: AGENTS.md rule 4 rewritten ("balanced **and** anchorable"), module-map lines for `region.ts`/`tools.ts` updated, historical reference in docs/media-visibility-design.md renamed.
- Test count matches the description: 4 new regression tests + 3 stale pins rewritten, including the issue #38 non-monotonic checkpoint shape `[8, 2..6, 7]`, where a mid-round request now resolves by stopping at the multi-call node instead of being rejected.
- One edge case I analyzed and accepted: with a spliced checkpoint strictly inside a requested span, the tier-3 walk could return the checkpoint's ref. That class (a kernel range spanning an interior checkpoint message) is already reachable today through plain bare-ref edges, so this PR introduces no new semantic — and strictly improves on the previous hard failure.

**Duplicate note**
PR #156 (lxckck) fixes the same root cause but is strictly narrower: 7 files vs 10, and its region.ts change does not touch `seqOfKernelRef` — so it misses the third spot (sub-id-bounded ranges vanishing from the nudge table) — plus fewer tests and no AGENTS.md rule rewrite. Suggest merging #157 and closing #156 as the duplicate. (Flagging per one-issue-one-PR; closing is yours to do.)

**Verdict**: the diff is clean — every line serves the fix — behavior verified end-to-end locally, regression coverage in place, docs in sync. Mergeable.

中文摘要：修复了并行多工具调用轮次 + 空工具结果构成的永久不可压缩死区（边界谓词把「无裸 ref」误判为「不可作边界」），让多调用消息经子 id、空结果经配对平衡直接充当压缩边界，并补上 nudge 范围表丢失子 id 边界行的第三处隐患；我已独立复验（typecheck / 334 项测试 / 构建全绿）并核对了文档同步与重复 PR（#156 范围更窄，缺 seqOfKernelRef 修复），本 PR 可以合并。

PR #160, opened 2026-09-20
(feat) rename the human slash command /acp to /acp-prune

### LINKED ISSUE #154: Slash command 命名与ACP协议容易产生混淆，建议改一下吧
如标题，虽说不是什么大问题吧，但总让人感觉很别扭，未来如果和某些ACP相关插件冲突了也会很尴尬不是么？

ranxianglei:
## Problem

The human slash command was registered under the name `acp`. That collides with what "acp" already means inside DSH — the **Agent Client Protocol** (`packages/acp` in the host). Two different things sharing one short name is confusing on its own, and it creates a real conflict risk: DSH's command registry is a `NamedEntries` map that throws `command "acp" is already registered` when two entries share a name, so any future ACP-related plugin that registers its own `acp` command would fail to register. A bare short name is exactly the shape most likely to be grabbed by another package.

## Cause

`src/commands.ts` registered the command with `name: 'acp'`. The name was inherited from upstream billion-context-pi, but we have long since diverged structurally (upstream splits this into several standalone commands; here it is one command with four subcommands), so keeping the upstream name buys nothing while costing the collision above.

## Fix

Rename the human slash command `acp` → `acp-prune` (Active Context Pruning). Clean switch — DSH has no alias mechanism, so the old `/acp` simply stops working (registering both would just re-introduce the collision surface this change removes).

Changed:
- `src/commands.ts`: registration `name: 'acp'` → `'acp-prune'`, plus every user-facing usage/error string and comment.
- Other `src/` files: usage strings and comments that reference the command.
- `tests/` and `scripts/e2e/run-e2e.mjs`: updated to the new name; added a regression test asserting the registered command name is `acp-prune` (and not `acp`) so a future slip-back fails loudly.
- `README.md`, `README.en.md`, `docs/`, `AGENTS.md`: kept in sync.

Deliberately unchanged: the model tool names (`compress` / `decompress` / `search_context` / `acp_status`), kernel/prompts copy, the `acp-kernel` package name, the `compaction-acp` composition id and settings namespace, plugin source markers, and the uppercase `ACP` abbreviation. Only the slash-command spelling changed.

## Verification

- `npm run typecheck` — green
- `npm test` — 331 tests pass, 0 fail
- `npm run build` — green
- `npm run test:e2e` — green
- `grep -rnF "/acp" src tests scripts` leaves only the new `/acp-prune` occurrences and the `acp_status` model-tool token (the two expected, explainable remainders).

## ⚠️ Breaking change

**The human slash command `/acp` is renamed to `/acp-prune`; the old name no longer works** (DSH has no alias mechanism, so there is no backward-compat shim). Model tools are unaffected. This should be called out explicitly in the next release notes.

Closes #154


ranxianglei:

[bot] 🏷 **Review verdict: diff is clean — one blocker found and fixed directly on the PR branch.**

**Diff review (first gate).** 32 files, +216/−208, every changed line is part of the rename:
- `src/`: registration `name: 'acp'` → `'acp-prune'` plus user-facing usage/error strings and comments only — zero logic changes.
- `tests/commands.test.ts`: the new regression test asserting the registered command name is `acp-prune` (and not `acp`) — good, this makes any future slip-back fail loudly.
- `scripts/e2e/run-e2e.mjs`, `README.md`, `README.en.md`, `docs/`, `AGENTS.md`, `cordis.patch.yml`: pure renames (verified line-by-line on the doc diffs). No unrelated files, no whitespace churn, no generated artifacts. Scope matches the description exactly, including everything deliberately unchanged (model tools, kernel copy, `acp-kernel` package name, `compaction-acp` id).

**Blocker found: the branch was behind main, so the PR could not merge as-is.** It was based on v0.2.23 (`4ff9a85`) while main had moved on (`#157` multi-call range fix + v0.2.24 release), and main advanced twice more during this review session (`#159` checkpoint-span test pin, `#161` tracker gate docs). Fixed directly on the PR branch:
- Rebased onto the current main tip `62266c69` (#161).
- Resolved two `docs/upstream-tracker.md` conflicts (the file moved under us both times): kept #161's restructured `#159` row and the new 「未解除的门」 section verbatim, and applied the new `/acp-prune status` spelling to the `#46` row. `AGENTS.md` auto-merged cleanly both times.
- Force-pushed to `ranxianglei:2026-09-20_rename-acp-slash-command` (head now `fd2a676`).

**Verification on the final head `fd2a676` (all run locally):**
- `npm run typecheck` — green
- `npm test` — 339/339 pass, 0 fail
- `npm run build` — green
- `npm run test:e2e` — PASS (incl. the renamed `human-side /acp-prune` check line)
- Stray-name sweep: `grep -rnF '/acp' src tests scripts docs README* AGENTS.md cordis.patch.yml` minus `/acp-prune` leaves only the expected remainders — the `acp_status` model-tool token, the `acpStatus` prompt-key identifier, `acp-kernel` package paths, DSH's own `packages/acp` (Agent Client Protocol) references, and pre-existing file/branch names. Matches your verification claim.

GitHub now reports the PR **mergeable: clean** against base `62266c69` (== current main tip; the branch parent is exactly that commit, so it fast-forwards). CI will run on the new head. One note: main moved three times during this ~1h review window — if it moves again before you merge, a fresh `git rebase origin/main` is a 30-second job (no code overlap with these docs commits).

The ⚠️ breaking-change callout in the PR body is already there for the next release notes — nothing needed from me on that side.

---
中文摘要:PR diff 干净(纯改名 + 新增命令名回归测试);唯一阻塞是分支落后 main(v0.2.24/#157,期间 #159、#161 又先后合入),已直接在 PR 分支上 rebase 到最新 main 并解决 `docs/upstream-tracker.md` 冲突(保留 #161 的重构内容、`#46` 行改用新名),本地全量验证通过(typecheck / 339 项测试 / build / e2e),当前对最新 main 可干净合并。

